PR #8, opened 2017-06-21
[src/lib.rs] Replace hardcode links with docblocks for pub extern crate

behnam:


behnam:
Looks like `cargo doc` doesn't handle docblocks for `pub extern crate` items. So, let's fix that before landing this.

screen shot 2017-06-21 at 2 08 37 pm

behnam:
Filed a bug report: https://github.com/rust-lang/rust/issues/42810

behnam:
Let's close this until the feature is supported in rustdoc.

PR #27, opened 2017-07-13
[ucd/normal] Add Conformance Test

### LINKED ISSUE #18: Add Conformance Test to ucd-normal component
Use the *extracted* data file for `DecompositionType` property to test conformance of the implementation.

Test source file:
http://www.unicode.org/Public/UCD/latest/ucd/extracted/DerivedDecompositionType.txt

NOTE: The test should also verify that the characters **not** listed in the file get `None` value for the property.

Similar conformance test code: https://github.com/behnam/rust-unic/blob/master/components/bidi/tests/conformance_tests.rs

CAD97:
Closes #18

... and it seems there's a test failure.

```
11172 test cases failed! (1100892 passed) {

0: Fail { line_num: Some(234), char: '가', exp_dt: Some(Canonical), actual_dt: None }

...

11171: Fail { line_num: Some(234), char: '힣', exp_dt: Some(Canonical), actual_dt: None }

}
```

Relevant line from [DecompositionType.txt]:

```
AC00..D7A3    ; Canonical # Lo [11172] HANGUL SYLLABLE GA..HANGUL SYLLABLE HIH
```

[DecompositionType.txt]: http://www.unicode.org/Public/UCD/latest/ucd/extracted/DerivedDecompositionType.txt

### DIFF
--- a/components/ucd/normal/tests/conformance_tests.rs
+++ b/components/ucd/normal/tests/conformance_tests.rs
@@ -31,6 +31,7 @@ struct Fail {
 }
 
 #[test]
+#[should_panic(expected = "11172 test cases failed! (1100892 passed)")]
 fn test_decomposition_type_conformance() {
     let mut passed_num: u32 = 0;
     let mut failures: Vec<Fail> = Vec::new();


CAD97:
Tests are going to fail because the added test fails (as far as I can tell, rightfully so).

components/ucd/normal/tests/unicode_version_tests.rs
```diff
@@ -20,3 +21,149 @@ fn test_version_against_ucd_core() {
         unic_ucd_core::UNICODE_VERSION
     );
 }
+
```
behnam: `tests/unicode_version_tests.rs` files are only to do checks on Unicode Version matters.

We need to move this test to a new file, like `tests/conformance_tests.rs`, similar to [`ucd-bidi` conformance tests](https://github.com/behnam/rust-unic/blob/master/components/bidi/tests/conformance_tests.rs).

components/ucd/normal/tests/unicode_version_tests.rs
```diff
@@ -20,3 +21,149 @@ fn test_version_against_ucd_core() {
...
+        // Data lines
+        {
+            let fields: Vec<_> = line.split(';').map(str::trim).collect();
+            let input_range = {
+                let range_text = fields[0];
+                if let Some(range_idx) = range_text.find("..") {
+                    let low_text = &range_text[0..range_idx];
+                    let high_text = &range_text[range_idx + 2..];
+                    let low =
+                        u32::from_str_radix(low_text, 16).expect("Parse error on base 16 string");
```
behnam: In tests, it's fine to just `unwrap()` and not explain the failure reason for obvious situations like this. And it can make these lines shorter.

components/ucd/normal/tests/unicode_version_tests.rs
```diff
@@ -20,3 +21,149 @@ fn test_version_against_ucd_core() {
...
+        if line.starts_with('@') {
+            panic!("State setting line in DecompositionTypeTest.txt; expected none");
+        }
+
+        // Data lines
+        {
+            let fields: Vec<_> = line.split(';').map(str::trim).collect();
+            let input_range = {
+                let range_text = fields[0];
+                if let Some(range_idx) = range_text.find("..") {
```
behnam: Also an alternate approach: if you use `str.split()`, it will give you an iterator and you won't need to range-index the sections out of the string.

components/ucd/normal/tests/unicode_version_tests.rs
CAD97: Whoops, my bad. If you want to preserve the state of `tests/unicode_version_tests.rs`, I can (look up how to and) squish the history of the PR to one commit. For now I'm just adding in another commit to move the code.

I'm doing this first, so it may break the review integration a little, but ¯\\\_(ツ)_/¯ I'm addressing the points anyway.

components/ucd/normal/tests/unicode_version_tests.rs
CAD97: Yeah, this turned out better, and seems more idiomatic with `Option` combinators.

CAD97:
didn't add the expect failure, one moment...

CAD97:
Unless I messed something up, this should have the requested changes and pass.

components/ucd/normal/tests/conformance_tests.rs
```diff
@@ -0,0 +1,153 @@
...
+            let fields: Vec<_> = line.split(';').map(str::trim).collect();
+            let input_range = {
+                let range_text = fields[0];
+                let mut range = range_text.split("..");
+                let low = u32::from_str_radix(range.next().unwrap(), 16).unwrap();
+                let high = range
+                    .next()
+                    .map(|c| u32::from_str_radix(c, 16).unwrap())
+                    .unwrap_or(low);
+                low..(high + 1)
```
CAD97: I just wish `..=` were stable.... It's ugly (imho, I also dislike every other solution, including `...`), but it's better than this.

components/ucd/normal/tests/conformance_tests.rs
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+        // State setting lines
+        if line.starts_with('@') {
+            panic!("State setting line in DecompositionTypeTest.txt; expected none");
+        }
+
+        // Data lines
+        {
+            let fields: Vec<_> = line.split(';').map(str::trim).collect();
```
CAD97: I could drop the `collect` call here and use `.next().unwrap()` like I did for the other split. In that case I think it makes sense in order to use `unwrap_or`, but it's open to debate here. I don't really have a strong opinion either way, other than avoiding the `collect` _may_ be marginally more performant and avoid a heap allocation.

components/ucd/normal/tests/conformance_tests.rs
```diff
@@ -0,0 +1,153 @@
...
+            failures.len(),
+            passed_num,
+            failures[0],
+            failures.len() - 1,
+            failures[failures.len() - 1],
+        );
+    }
+}
+
+fn get_dt_from_name(name: &str) -> DT {
```
CAD97: Should I add a test for this fn?

Also,
- Should the short forms be included? I don't think they're used in the source txt (but I haven't read the whole thing).
- Should this be part of the enum instead of a helper here?

components/ucd/normal/tests/conformance_tests.rs
behnam: Yeah, probably would get better perf if not using `collect()` and use `nth()` to access. But, since it's just a test and not part of the lib, it's fine either way.

components/ucd/normal/tests/conformance_tests.rs
behnam: We still don't have a common API for accessors for UCD-defined long/abbreviated value names. So, it's fine to keep it here for now, and we will move it the original class, or at least have a reverse test for it later.

components/ucd/normal/tests/conformance_tests.rs
```diff
@@ -0,0 +1,153 @@
...
+// except according to those terms.
+
+
+extern crate unic_ucd_normal;
+extern crate unic_ucd_utils;
+
+use unic_ucd_normal::DecompositionType as DT;
+use unic_ucd_utils::iter_all_chars;
+use std::collections::HashSet as Set;
+use std::{char, u32};
```
behnam: IIRC, this line is not needed anymore.

components/ucd/normal/tests/conformance_tests.rs
```diff
@@ -0,0 +1,153 @@
...
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+
+extern crate unic_ucd_normal;
+extern crate unic_ucd_utils;
+
+use unic_ucd_normal::DecompositionType as DT;
+use unic_ucd_utils::iter_all_chars;
+use std::collections::HashSet as Set;
```
behnam: In UNIC, I'm trying to keep `use` statements in four groups, split with one empty line: 1) core/std lib, 2) third-party, 3) this workspace, and 4) this package. (I should write the style guide soon, it's not anywhere yet.)

And another nit: usually we don't rename unless it's useful or necessary. Using original name, `HashSet`, would be fine here.

components/ucd/normal/tests/conformance_tests.rs
```diff
@@ -0,0 +1,153 @@
...
+#[test]
+#[should_panic(expected = "11172 test cases failed! (1100892 passed)")]
+fn test_decomposition_type_conformance() {
+    let mut passed_num: u32 = 0;
+    let mut failures: Vec<Fail> = Vec::new();
+    let mut to_consider: Set<char> = iter_all_chars().collect();
+
+    for (line_idx, line) in DT_TEST_DATA.lines().enumerate() {
+        // Remove comments and trailing whitespace
+        let line = if let Some(idx) = line.find('#') {
```
behnam: `split()` would make this one easier, too, I think.

PR #28, opened 2017-07-13
cargo fmt --all

CAD97:
This PR is the format changes suggested by the latest release of rustfmt (`rustfmt-nightly:0.1.9`).

This is a suggestion to update to that style, rather than that of the old Syntex-based rustfmt (`rusftmt:0.9.0`) which has been discontinued due to Syntex's being discontinued, in favor of using libsyntax.

Original text below.

<hr />

Because this is a monorepo with multiple Cargo.toml, `cargo +nightly fmt` doesn't format sub-crates. (If there is functionality to do so, I haven't found it, and it should likely be added.)

In order to format a subcrate you must `cd` to the directory which contains the Cargo.toml describing the crate you are working on and then run rustfmt.

In order to keep code formatted well it must first be formatted well. I don't take hitting myself with the `git blame` on so many lines lightly, but having the code formatted by the guides laid out in `.rustfmt.toml` is important, and allows future contributors to run rustfmt without formatting code they didn't change.

If there's a format change you don't want here, the answer is to tell rustfmt to format it differently; by including `.rustfmt.toml` you're encouraging contributors to rustfmt their code and that is the style guide of the repository.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 *.rs.rustfmt
 target/
 Cargo.lock
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@
 *.rs.rustfmt
 target/
 Cargo.lock
-.idea/
--- a/components/bidi/src/explicit.rs
+++ b/components/bidi/src/explicit.rs
@@ -100,7 +100,10 @@ pub fn compute(
                         // Pop everything up to and including the last Isolate status.
                         match stack.vec.pop() {
                             None |
-                            Some(Status { status: OverrideStatus::Isolate, .. }) => break,
+                            Some(Status {
+                                status: OverrideStatus::Isolate,
+                                ..
+                            }) => break,
                             _ => continue,
                         }
                     }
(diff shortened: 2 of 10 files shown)


CAD97:
If you decide to take this it would also be nice to [enforce the formatting](https://github.com/rust-lang-nursery/rustfmt#checking-style-on-a-ci-server).

behnam:
Thanks for looking into these, @CAD97 .

Regarding current status, the code is already fully formatted with a recent `rustfmt-nightly` version.

For cargo workspaces (like UNIC), many cargo commands take a `--all` option to work on all the packages in the workspace. So, you need to do `cargo +nightly fmt --all`. In this repo, we have a script `.cargo/format.sh` that does that, plus formatting the Python code.

To put the formatting as a blocker for merging, I haven't figured out yet if it's a good thing to have or not. Many other repos are still waiting for a more stable `rustfmt` to do so. Maybe we can try it and see if it slows us down or not.

CAD97:
These changes are all from running rustfmt. So if the code was already formatted, then I guess rustfmt changed, so that would be a point for not blocking on it...

I think Travis supports a non-blocking breakage anyway (recommended for nightly in the example setup), so maybe stick the format in there? That way Travis could report it without blocking, and the reviewer (or just user) could then make an informed decision.

behnam:
Hum... I'm on `rustfmt:0.9.0-nightly` and don't get these changes running `cargo +nightly fmt --all`. What version are you on?

CAD97:
I installed the wrong one.... (`rustfmt instead of rustfmt-nightly`)

Whoops.

CAD97:
Actually, no, never mind.

[rustfmt-nightly](https://crates.io/crates/rustfmt-nightly) is 0.1.9.
[rustfmt](https://crates.io/crates/rustfmt) is 0.9.0.

[They changed crates.io artifacts](https://users.rust-lang.org/t/rustfmt-releases/11357) when moving away from Syntex to libsyntax.

behnam:
Right, @CAD97. Rustfmt has got updates. Thanks! :)

PR #31, opened 2017-07-14
[ucd/category] Provisionally add unic-ucd-category

CAD97:
My use cases mainly draw from [DerivedCoreProperties], one of which the main builders for is General_Category. And the best way to learn is to dive right in, so:

This updates `gen_ucd_tables.py` and adds the crate with enum for ucd-category. The design was mostly ad-hoc based on other crates in UNIC, but I expect that something can be improved. At the very least, my placeholder version `0.0.1` will probably be changed, and unic-ucd and unic bumped as well.

I skipped over name in the [list of planned ucd crates][ucd-crates] because there are more interesting compression opportunites available there, such as a prefix trie or such, and I'm not quite prepared to dive into how [other crates][unicode_names] have done it quite yet.

[DerivedCoreProperties]: http://unicode.org/reports/tr44/#DerivedCoreProperties.txt
[ucd-crates]: https://github.com/behnam/rust-unic/tree/master/components/ucd
[unicode_names]: https://crates.io/crates/unicode_names


components/ucd/category/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+    /// Unassigned
+    ///
+    /// Abbreviated: Cn
+    Unassigned,
+}
+
+
+use self::GeneralCategory::*;
+
+// TODO: potentially change these to match pattern matching.
```
CAD97: I'd love to see a comparison here. There's a large possibility that the two could be optimized to the same thing or that the pattern match is more performant. I feel that this form is much clearer; for comparison:

```rust
fn is_cased_letter(&self) -> bool {
    [UppercaseLetter, LowercaseLetter, TitlecaseLetter].contains(self)
}
```

versus

```rust
fn is_cased_letter(&self) -> bool {
    match *self {
        UppercaseLetter | LowercaseLetter | TitlecaseLetter => true,
        _ => false,
    }
}
```

There's an even bigger difference when more symbols are used, such as in `is_punctuation`.

components/ucd/category/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+// Static array contains was easier and clearer to get the first draft up,
+// but is it optimized to the same thing?
+impl GeneralCategory {
+    /// Lu | Ll | Lt
+    ///
+    /// Abbreviated: LC
```
CAD97: These comments are probably close to useless, but they're how [tr44 describes them](http://unicode.org/reports/tr44/#General_Category_Values).

components/ucd/category/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+    include!("tables/general_category.rsv");
+
+impl GeneralCategory {
+    /// Find the GeneralCategory of a single char.
+    pub fn of(ch: char) -> GeneralCategory {
+        bsearch_range_value_table(ch, GENERAL_CATEGORY_TABLE)
+    }
+}
+
+fn bsearch_range_value_table(
```
CAD97: This fn is copied almost verbaitim from `ucd/normal/decomposition_type.rs`. Is it a good candidate for extracting into a helper? I suspect something like a `ucd-util` that contains the most commonly used impls for searching the commonly used table formats would be useful.

components/ucd/category/tests/unicode_version_tests.rs
```diff
@@ -0,0 +1,22 @@
...
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+
+extern crate unic_ucd_core;
+extern crate unic_ucd_category;
+
+
+//#[test]
```
CAD97: Whoops, need to re-enable this (had it disabled before the python build script was edited).

tools/gen_ucd_tables.py
```diff
@@ -179,23 +179,23 @@ def get_bidi_class_info():
...
         # Cypriot_Syllabary, Phoenician, Lydian, Meroitic Hieroglyphs, Meroitic
         # Cursive, Kharoshthi, others
         (0x10800, 0x10FFF, "R"),
-        (0x1E800, 0x1EDFF, "R"),    # others
-        (0x1EF00, 0x1EFFF, "R"),    # others
+        (0x1E800, 0x1EDFF, "R"),  # others
+        (0x1EF00, 0x1EFFF, "R"),  # others
         # default to ET
-        (0x20A0, 0x20CF, "ET"),     # Currency Symbols
+        (0x20A0, 0x20CF, "ET"),  # Currency Symbols
```
CAD97: These format changes must have been done automatically by my ide (bad IDE! only touch VCS changed text unless told otherwise!); I can change them back, and probably should.

tools/gen_ucd_tables.py
```diff
@@ -487,6 +487,80 @@ def emit_normal_tests_tables(dir):
...
+    return (
+        general_category_mapping
+    )
+
+
+def emit_general_category_tables(dir):
+    general_category_mapping = get_general_category_mapping()
+
+    with open(join(dir, 'general_category.rsv'), "w") as values_file:
+        rustout.emit_table(
```
CAD97: This variation of `rustout.emit_table` feels like it will be used a lot for the simple properties; would it be worth making a version with this default `print_fun`?

components/ucd/category/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "unic-ucd-category"
+version = "0.0.1"
+authors = ["The UNIC Project Developers"]
+homepage = "https://github.com/behnam/rust-unic/"
+repository = "https://github.com/behnam/rust-unic/"
+license = "MIT/Apache-2.0"
+keywords = ["text", "unicode"]
+description = "UNIC - Unicode Character Database"
```
behnam: Don't forget to update the `description` field.

components/ucd/category/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+readme = "README.md"
```
behnam: Drop this if we don't have a `README.md` on this crate. (The small crates don't have one yet.)

components/ucd/category/src/lib.rs
```diff
@@ -0,0 +1,313 @@
+// Copyright 2012-2015 The Rust Project Developers.
```
behnam: This line is only needed ifyou are copying content from the Rust project, in which case should have the same years as the original file. Is this the case here?

components/ucd/category/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+
+#![deny(unsafe_code, missing_docs)]
+
+//! # UNIC — UCD — Category
+//!
+//! A component of [`unic`: Unicode and Internationalization Crates for Rust](/unic/).
+//!
+//! Unicode General Category.
+//!
+//!
```
behnam: This is the first thing people see on the package page on crates.io, and it's a very basic and useful character property. Would be great to have a bit more text here about it. 

And don't forget to link to the definition of the property on unicode.org. (See links on other packages for formatting.)

components/ucd/category/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+
+/// The [Unicode version](http://www.unicode.org/versions/) of data
+pub const UNICODE_VERSION: UnicodeVersion = include!("tables/unicode_version.rsv");
+
+/// Represents the Unicode Character
+/// [*General Category*](http://unicode.org/reports/tr44/#General_Category) property.
+///
+/// * <http://unicode.org/reports/tr44/#General_Category_Values>
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum GeneralCategory {
```
behnam: I recently started putting all type-level abstraction in their own module. So, `Age` enum and its various implementations and tests are in `age.rs`. And, `char`/`str` helper traits go into `traits.rs`, and only a couple of needed things get `pub use`d here in `lib.rs` from different modules.

It has shown to be a good way to keep unrelated matters separate. Would be great to do the same here.

PS. I'll start writing these in `CONTRIBUTING.md`. (https://github.com/behnam/rust-unic/issues/36)

components/ucd/category/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+    /// An uppercase letter
+    ///
+    /// Abbreviated: Lu
```
behnam: Too bad Rust's `enum`s don't get *alternate* names. We could really use them here in UCD packages.

Since it's markdown and will show up in docs, let use backticks for abbreviated names.

Also, you can `cargo doc` the package and check out the output and see if having a two-paragraph doc is okay here, or we better change it to one line, like:
```
An uppercase letter (Abbreviated name: `Lu`)
```

components/ucd/category/src/lib.rs
behnam: The later is kind of guaranteed to be the fastest in Rust. The former may optimize to the same performance, but that's taking a chance.

If you don't like the second form, then maybe this is a good time that we start using `matches` crate. It gives you the `match` feature in a macro call. It's a very stable little package, so is fine to depend on.

components/ucd/category/src/lib.rs
behnam: Not useless at all! Very good for documentation! And maybe one day we find a way to let people use the shorter names alongside the longer ones.

components/ucd/category/src/lib.rs
behnam: Well, that's the plan. If you want to put them in the utils crates, please do so in a different diff to make it easier to review.

That said, we're thinking of replacing the python code with <https://github.com/BurntSushi/rucd>, so we can use other Rust packages during table generation, like <https://github.com/behnam/rust-phf>. That's why I haven't spent much time on the cleaning up the current table code helpers.

components/ucd/category/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+        }
+        assert_eq!(GC::of(' '), GC::SpaceSeparator);
+        assert_eq!(GC::of('!'), GC::OtherPunctuation);
+        assert_eq!(GC::of('"'), GC::OtherPunctuation);
+        assert_eq!(GC::of('#'), GC::OtherPunctuation);
+        assert_eq!(GC::of('$'), GC::CurrencySymbol);
+        assert_eq!(GC::of('%'), GC::OtherPunctuation);
+        assert_eq!(GC::of('&'), GC::OtherPunctuation);
+        assert_eq!(GC::of('\''), GC::OtherPunctuation);
+        // TODO I would like to do the rest of ASCII
```
behnam: Yeah, corner cases of ASCII would be great, like U+0000, U+007F, and also edges of BMP (U+FFFF, U+FFFE, U+FFEF), and edges of PUA planes, specially (U+10FFFF).

components/ucd/category/tests/unicode_version_tests.rs
behnam: 👍

tools/gen_ucd_tables.py
```diff
@@ -28,12 +28,12 @@
 from common import path, memoize
 from unicode_utils import is_surrogate
 
-
```
behnam: nit: Python's PEP-8 recommends two empty lines here, so would be great to keep it.

You can just run the `.cargo/format.sh` script to get it fixed if you have the CLI tool installed.

tools/gen_ucd_tables.py
behnam: Yeah, would be great to have them in one column. Don't worry if the IDE is bothering too much.

tools/gen_ucd_tables.py
behnam: Fine either way. (See my comment above about dropping the Python code completely.)

components/ucd/Cargo.toml
```diff
@@ -18,3 +18,4 @@ unic-ucd-bidi = { path = "bidi/", version = "0.4.0" }
 unic-ucd-core = { path = "core/", version = "0.4.0" }
 unic-ucd-normal = { path = "normal/", version = "0.4.0" }
 unic-ucd-utils = { path = "utils/", version = "0.4.0" }
+unic-ucd-category = { path = "category/", version = "0.0.1" }
```
behnam: Right. I think you can start with `0.4.0`. There won't be a release of it, but it allows the code work and we will update it to `0.5.0` for the next general release.

components/ucd/category/src/lib.rs
CAD97: I'm actually experimenting with porting the python scripts into Rust (to use in cargo buildscripts), partially because I'm just interested, and partially because it could potentially work. And I'm enjoying it (lets me exercise understanding Python code and get more practice writing Rust, so it's not wasted effort. And it's a problem space I wanted to dip my toes into.

Assume rucd works for the generation, it'd definitely be best to use that, once its ready.

components/ucd/category/src/lib.rs
CAD97: `matches` has a very nice representation. I've pulled it in for this sub-crate.

components/ucd/category/src/lib.rs
CAD97: `use UppercaseLetter as Lu`? In any case it's probably best to "force" use of the descriptive names.

components/ucd/category/src/lib.rs
CAD97: Note there's one really big small problem with using PHF macros: https://github.com/rust-lang/cargo/issues/1359

PHF is _really slow_ without release optimizations for large maps.

Other than that, it'd definitely be a plus to use a PHF map where it would be appropriate.

components/ucd/category/src/lib.rs
CAD97: Whoops, I don't actually know how that sneaked in.

components/ucd/category/src/lib.rs
CAD97: Yeah, I think I've seen that guideline in a few places (and mostly agree with it): `lib.rs` should really just be a list of `[pub] mod` and `pub use`.

components/ucd/category/src/lib.rs
CAD97: I agree, this looks better. I switched to calling it the short form because Unicode does.

tools/gen_ucd_tables.py
CAD97: Didn't have autopep8 installed previously (don't write too much python and PyCharm/IDEA usually do a decent job of formatting along pep8). Now I do.

Strangely, it's not changing this. I went ahead and changed it back manually.

components/ucd/Cargo.toml
CAD97: 👌

CAD97:
👌 we should be good now!

CAD97:
Wait, I need to add myself to `update_ucd_data.py` don't I?

(And I'm pretty sure we should be fetching from http://www.unicode.org/Public/10.0.0/ucd/ and not http://www.unicode.org/Public/UCD/latest/ucd/

behnam:
> (And I'm pretty sure we should be fetching from http://www.unicode.org/Public/10.0.0/ucd/ and not http://www.unicode.org/Public/UCD/latest/ucd/ )

This is another issue we need to address sometime soon. I want the update script to take the version of unicode to download the files for, instead of always getting the latest. The same for the IDNA update script.

PR #32, opened 2017-07-14
Move test dependencies into [dev-dependencies]

CAD97:
A simple bookkeeping change.

No version bumps were made. No code or API was changed, just the explicit dependency chain.


behnam:
Awesome! Thanks!

Btw, this looks like something that a linter/formating tool can do. I don't know if there exist any Cargo-level linter/formatting tool besides `rustfmt`. If there's any, we can start using it.

CAD97:
[I suggested a lint for clippy.](https://github.com/Manishearth/rust-clippy/issues/1892)

PR #38, opened 2017-07-16
Rust build pipeline

CAD97:
DO NOT MERGE YET!

This is to track and get feedback on my experimental port of the python buildscripts into Rust.

This will allow us to use utilities such as PHF to generate our `rsv` tables.


CAD97:
`cargo run --package=unic-gen` now cleans out all the `tables` directories and generates only ` components/ucd/core/src/tables/unicode_version.rsv`.

The build will be broken for a bit because I accidentally committed the deleting of the other `rsv`.

But this proves the system works! (Though I still need to check `--update`.)

CAD97:
At this point all non lookup table parts of the buildscript have been moved into Rust. I've deleted what I'm sure is replaced by the Rust code from the Python script.

I'm going to take a short hiatus from working on this direct port to explore a more rewritten build pipeline now that I understand the current one better.

CAD97:
Supplanted by #47

Issue #37, opened 2017-07-16
[ucd/bidi] BidiClass should use the long names


CAD97:
The [Bidi_Class_Values] enum should use the long names of the bidi classes, for clarity and to fit in better with the rest of the ucd api and the Rust ecosystem.

This can probably be bikeshedded ad nauseum, but defaulting to the descriptive names seems the better idea and [§5.8.1 Property Aliases] tells us that the long symbolic names are the preferred aliases. (Cases like `Age` where we can provide a more meaningful struct rather than a enum excepted, of course.)

We could offer a `alias` mod or such which provides `pub use` bindings to the abbreviated symbolic name. [PropertyValueAliases.txt] could (in theory) be used to generate and/or test the aliases.

[Bidi_Class_Values]: http://unicode.org/reports/tr44/#Bidi_Class_Values
[§5.8.1 Property Aliases]: http://unicode.org/reports/tr44/#PropertyValueAliases.txt
[PropertyValueAliases.txt]: http://www.unicode.org/Public/UCD/latest/ucd/PropertyValueAliases.txt

behnam:
Good point and great idea! Having a mod that only aliases the names sounds like the best alternate to allow access to the same enum variant under different names, while supporting `use ...::*` for both formats.

For the mod name, I like the term `abbr_names` better, because it makes it clear what kind of alias the names are, and leave it open to have other sets of aliases, if needed in any case.

Then, it's a matter of the mod path. Do we want to have `unic::ucd::BidiClass::abbr_names`, or `unic::ucd::bidi_class_abbr_names`, or `unic::ucd::bidi_class::abbr_names` (we don't have `unic::ucd::bidi_class` at the moment)?

I don't know if there are any similar cases in the current Rust code to see what's the common practice. So, we can try whatever that's more intuitive and practical to use, and clear to implement. I think the first path is the best in these perspectives.

What do you think?

CAD97:
If the first is valid, it makes the most sense. Even if it's not actually that way, it feels like the `abbr_names` namespace is a child of `BidiClass`.

I just tested that, and I couldn't get it to work.

Because it's exposed currently as `unic::ucd::BidiClass`, maybe have the "pseudo enum" as `unic::ucd::BidiClassAbbr`:

```rust
pub enum Enum {
    LongName1,
    LongName2,
}
pub mod EnumAbbr {
    pub use super::Enum::LongName1 as L1;
    pub use super::Enum::LongName2 as L2;
}
```

[...]

PR #39, opened 2017-07-17
Create bidi_class::abbr_names and use long names in BidiClass

behnam:
Update the rest of code-base to use the new enum variant names and
aliases.

Closes <https://github.com/behnam/rust-unic/issues/37>

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,11 +173,11 @@ fn main() {
     assert!(text.has_rtl());
     assert!(text.has_ltr());
 
-    assert_eq!(text.chars().nth(0).unwrap().bidi_class(), BidiClass::R);
+    assert_eq!(text.chars().nth(0).unwrap().bidi_class(), BidiClass::RightToLeft);
     assert!(!text.chars().nth(0).unwrap().is_ltr());
     assert!(text.chars().nth(0).unwrap().is_rtl());
 
-    assert_eq!(text.chars().nth(3).unwrap().bidi_class(), BidiClass::L);
+    assert_eq!(text.chars().nth(3).unwrap().bidi_class(), BidiClass::LeftToRight);
     assert!(text.chars().nth(3).unwrap().is_ltr());
     assert!(!text.chars().nth(3).unwrap().is_rtl());
 
--- a/components/bidi/src/bidi_info.rs
+++ b/components/bidi/src/bidi_info.rs
@@ -16,17 +16,18 @@ use std::fmt;
 use std::iter::repeat;
 use std::ops::Range;
 
+use unic_ucd_bidi::BidiClass;
+use unic_ucd_bidi::bidi_class::abbr_names::*;
+
 use explicit;
 use implicit;
 use level;
 use prepare;
 use format_chars;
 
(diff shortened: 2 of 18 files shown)


behnam:
This is a temp solution to use long names for `BidiClass` enum variants. Now, having this, and the other enum types, I think we can put the common parts in a macro to do something like this:

```rust
enum_char_property!(
    enum BidiClass {
        [...]

        LeftToRight {
            abbr_name => L,
            long_name => "Left_To_Right",
            description => "Left-to-Right",
        },

        LeftToRightEmbedding {
            abbr_name => LRE,
            long_name => "Left_To_Right_Embedding",
            description => "Left-to-Right Embedding",
        },

        [...]
    }
)
```

which sets enum variant doc using description, creates an `impl BidiClass` with methods `long_name(&self)`, `abbr_name(&self)`, `description(&self)`, and `display(&self)`.

What do you think, @CAD97 ?

CAD97:
The pattern looks good to me! And personally I like the more match-like syntax over `quick-error`'s approach.

Also a good thing; this allows us to add in short name bindings as associated consts to all enums with one change when that reaches stable, if we decide to do so.

As we bring this in, it might be worth trait-izing the different [property types](http://unicode.org/reports/tr44/#Type_Key_Table) that the UCD exposes:

- Catalog: properties with enumerated values which are expected to be regularly extended in successive versions of the Unicode Standard
- Enumeration: properties with enumerated values which constitute a logical partition space; new values will generally not be added to them in successive versions of the standard
- Binary: a special case of Enumeration properties, which have exactly two values: Yes and No (or True and False)
- String: mappings from a Unicode code point to another Unicode code point or sequence of Unicode code points; examples include case mappings and decomposition mappings
- Numeric: properties specify the actual numeric values for digits and other characters associated with numbers in some way
- Miscellaneous: properties that do not fit neatly into the other property categories

I'm not sure exactly what design would be best, but as an example,

```rust
trait EnumeratedEnumProperty : Display {
    fn long_name(&self) -> &'static str;
    fn abr_name(&self) -> &'static str;
    fn description(&self) -> &'static str;
}
```

The idea being to communicate more of the information the UCD gives us and to bundle up this related functionality into a trait.

Alternatively, just have a `trait UCDProperty` which has those `fn`; I'm pretty sure every property has a long and a short name, and we want to provide a description/Display to all of them as well.

behnam:
Right, we need to create traits to better define common expectations from char-property types. But I'm not sure we need a one-to-one mapping between Unicode-defined types and the Rust code. I think we can let it grow organically and see what's needed, because we don't exactly have a one-to-one mapping for the implementation.

For example, `Age` is expected to be a plain *Catalog* type by Unicode, but we have a two-variant Rust `enum` type for it, `{ Assigned(_), Unassigned }`, of which the first one also holds a `UnicodeVersion` value for the actual `Age` value. So, in Rust, it's a mix of Enum and Numeric.

On another hand, we have `Script` and `Block`, which have to be plain Rust `enum`s, while being a Unicode *Category* type.

Okay, so let's land this, and we will work on the traits soon, covering all properties already implemented, and expand it from there

PR #41, opened 2017-07-19
Char property macro prototype 1

CAD97:
This PR provides a macro to somewhat automate the implementation of Unicode `Catalog` and `Enumeration` property types (and uses it in ucd/bidi and ucd/category).

I haven't touched ucd/normal because I have yet to do the necessary reading about the composition/decomposition specification in Unicode that is a prerequisite to me trusting myself to make changes in it.

Open questions:

- Should this be in `unic-ucd-core`? I'm starting to think so; every UCD crate will almost certainly want to use the macro if they expose a Catalog or Enumeration property. And since it's a macro, they just wouldn't `#[macro_use]` on their import (which they need for `UnicodeVersion`).
- Should we provide a `try_from_name(&str) -> Option<$name>` (or similar) fn? To do so would entail adding the `Long_Name` form to the macro definition. (It's just an `$ident` and syntax bikeshedding away from being added.)
- Any further macro syntax bikeshedding. I chose the current one for implementation because it's simple and doesn't run into the problem of wanting to have optional or reordered "members" of the variant (which my macro-fu is not good enough to support yet).
- The mythical `UnicodeProperty` trait. It's a seperate issue to this macro, but related. Since `Number` properties exist, I'm thinking the only across-all-properties functionality we can promise is an `of(ch: char) -> Self`. Should we also have a `UnicodePropertyWithAbbreviatedName` <sub><sup>(horrible example name)</sup></sub> that exposes `abbr_name(&self)` and `name(&self)`? In any case, since they're all constructed the same way with the same methods they should _probably_ be promised in a trait, even if they remain intrinsic methods.

### DIFF
--- a/components/ucd/bidi/Cargo.toml
+++ b/components/ucd/bidi/Cargo.toml
@@ -13,3 +13,4 @@ travis-ci = { repository = "behnam/rust-unic", branch = "master" }
 
 [dependencies]
 unic-ucd-core = { path = "../core/", version = "0.4.0" }
+unic-ucd-utils = { path = "../utils/", version = "0.4.0" }
--- a/components/ucd/bidi/src/bidi_class.rs
+++ b/components/ucd/bidi/src/bidi_class.rs
@@ -10,74 +10,162 @@
 // except according to those terms.
 
 use std::cmp::Ordering;
-use std::fmt;
 
 
-/// Represents the Unicode character
-/// [*Bidi_Class*](http://www.unicode.org/reports/tr44/#Bidi_Class) property, also known as the
-/// *bidirectional character type*.
-///
-/// * <http://www.unicode.org/reports/tr9/#Bidirectional_Character_Types>
-/// * <http://www.unicode.org/reports/tr44/#Bidi_Class_Values>
-#[derive(Clone, Copy, Eq, PartialEq, Debug, Hash)]
-#[allow(missing_docs)]
-pub enum BidiClass {
-    ArabicLetter,
-    ArabicNumber,
-    ParagraphSeparator,
-    BoundaryNeutral,
-    CommonSeparator,
-    EuropeanNumber,
-    EuropeanSeparator,
-    EuropeanTerminator,
-    FirstStrongIsolate,
-    LeftToRight,
-    LeftToRightEmbedding,
-    LeftToRightIsolate,
-    LeftToRightOverride,
-    NonspacingMark,
-    OtherNeutral,
-    PopDirectionalFormat,
-    PopDirectionalIsolate,
-    RightToLeft,
-    RightToLeftEmbedding,
-    RightToLeftIsolate,
-    RightToLeftOverride,
-    SegmentSeparator,
-    WhiteSpace,
-    // [UNIC_UPDATE_ON_UNICODE_UPDATE] Source: `tables/bidi_class_type.rsv`
-}
+char_property! {
+    // TODO: Once 1.20 comes, add the rest of the enum variants' docs
+    /// Represents the Unicode character
+    /// [*Bidi_Class*](http://www.unicode.org/reports/tr44/#Bidi_Class) property, also known as the
+    /// *bidirectional character type*.
+    ///
+    /// * <http://www.unicode.org/reports/tr9/#Bidirectional_Character_Types>
+    /// * <http://www.unicode.org/reports/tr44/#Bidi_Class_Values>
+    pub enum BidiClass {
+        // == Strong == //
 
+        /// Any strong left-to-right character
+        // ///
+        // /// ***General Scope***
+        // ///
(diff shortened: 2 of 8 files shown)


CAD97:
-.- how did I miss my own tests...

behnam:
This is cool! Thanks for working on it, @CAD97. It was fast!

> I haven't touched ucd/normal because I have yet to do the necessary reading about the composition/decomposition specification in Unicode that is a prerequisite to me trusting myself to make changes in it.

I think there are a couple of helper functions in `ucd/normal/` that I think belong to `normal/`. Just FYI. We need to fix that when we get back there.

> Should this be in unic-ucd-core? I'm starting to think so; every UCD crate will almost certainly want to use the macro if they expose a Catalog or Enumeration property. And since it's a macro, they just wouldn't #[macro_use] on their import (which they need for UnicodeVersion).

I think this belongs to a non-ucd component, like `unic::utils::char_property`, because it has nothing to do with UCD data (and specially unicode-version, which is the base of `ucd::core`), and other components may want to use it without having anything to do with UCD. We don't have such a component yet, and IDNA, Emoji and other components with char-property still need some UCD access. But, it's still valid for external users.

So, I was thinking that any helper code, meaning not depending on unicode/i18n data and algorithm, should go under `unic::utils`.

> Should we provide a try_from_name(&str) -> Option<$name> (or similar) fn?

I think yes. For now, mainly because we usually need these for conformance tests.

> To do so would entail adding the Long_Name form to the macro definition. (It's just an $ident and syntax bikeshedding away from being added.)

I think there are a bunch of fields that need to be defined for each variant. That's why I initially suggested to have a named syntax, instead of an ordered syntax. And, having the names, many of the fields would be optional, and fall back to another one if not provided.

For example, there are property values with equal abbr and long names. If we want to provide one function to match both (which I think we should), we only need to list these once. Leaving `long_name` field empty in the variant definition would be a sign for this.

Also, maybe we can even auto-gen long names from variant names. There are a couple of small crates for this, and I think we can depend on them, or just implement/copy it from there under its own `utils` crate, in case we need a very small portion of it, or want to customize and improve.

> Any further macro syntax bikeshedding. I chose the current one for implementation because it's simple and doesn't run into the problem of wanting to have optional or reordered "members" of the variant (which my macro-fu is not good enough to support yet).

I think optional fields would be a great thing to have here, because otherwise a lot of manual formatting/etc will be needed for each variant.

If we keep the diff limited to implementing the macro and an integration test for it, we can land it and work on it while the rest of the code is evolving, and switch whenever we feel it's ready.

> The mythical UnicodeProperty trait. [...]

I want to work on these traits before we get the macro out, because, as you already mentioned, this sets the contract for the macro.

What I have in mind at the moment is one simple `trait UnicodeProperty`, which applies to all types of properties. And, then, have sub-traits like `UnicodeEnumeratedProperty`, which has to be `enum` and have lots of goodies. We can have `long_name()`, etc directly on `UnicodeEnumeratedProperty` until we find the sweet spot for a new trait that provide those things (and not be an `enum`).

Let me take a shot at this tomorrow and we talk on that PR.

WDYT?

CAD97:
You take a shot and see what you can get. The annotated members is probably a better design for the macro; I just lack the macro knowledge required for the munching required to allow the flexibility it implies.

I agree that adding the formal `UnicodeProperty` trait at the same time as this probably would make sense; would the trait live in `unic::utils` as well? I do think that the macro for implementing the trait should be alongside the trait it implements.

I'll make the simple changes you've suggested for a fair comparison, but I can't wait to see what you come up with. Just keep [this macro limitation](https://www.reddit.com/r/rust/comments/6o3tkc/identifying_a_betanightly_macro_change/?ref=share&ref_source=link) in mind; I've required exactly one attribute before each enum variant because of it, but you could also choose to require a sigil (such as `case` (probably, untested) or a single symbol, such as `@`) before each variant.

CAD97:
If it's possible to rewrite a bound `$ident` "`BidiClass`" to "`Bidi_Class`" (with `macro_rules!`), I can't find it; I don't think it's possible.

The reason for taking in the long name form is to allow using it as an `&'static str` rather than having to allocate space on the heap and transform a slightly different str into a correct String.

behnam:
> I agree that adding the formal UnicodeProperty trait at the same time as this probably would make sense; would the trait live in unic::utils as well? I do think that the macro for implementing the trait should be alongside the trait it implements.

Actually, I remembered this morning that I meant `CharProperty` trait, but not `UnicodeProperty`. I don't know if we ever want `UnicodeProperty`, since there's nothing special about Unicode-defined/UCD-defined properties comparing to other ones, like those defined in Unicode IDNA/Emoji specs, or the IETF-defined ones.

So, I think we can put both `trait CharProperty` and `macro char_property` in the same crate: `unic::utils::char_property`.

I'll submit a PR today.

CAD97:
Supplanted by #48

PR #42, opened 2017-07-19
[Cargo] Exclude tools, data and data-dependent tests/benches in packages

### LINKED ISSUE #34: Exclude tools, data and data-dependent tests/benches in packages
We should use the [Cargo `exclude`](http://doc.crates.io/manifest.html#the-exclude-and-include-fields-optional) option to not publish data files to crates.io, as they are usually very large and almost of no use for 99.99% of the callers.

See https://github.com/servo/unicode-bidi/issues/43 for report issue with publishing tests without data when published crate is being used in repackaging.

See also https://github.com/rust-lang/cargo/issues/4268 for current limits of the `exclude` syntax, and the changes coming.

CAD97:
Should close #34 

I'm not exactly certain how to reproduce a "download from Crates.io" locally to see if the package A) builds and B) tests successfully, but looking at the output of `cargo package --allow-dirty --list` looks optimistic.

The benefit of closing #34 will be to remove the `data/` directory from the root crate (currently ~21MiB) and allowing the subcrates to pass their tests locally without being in this repository layout.

If someone can test to make sure that those two things are true of the package, then this should be good to merge.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 *.rs.rustfmt
 target/
 Cargo.lock
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@
 *.rs.rustfmt
 target/
 Cargo.lock
-.idea/
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,13 @@ description = "UNIC - Unicode and Internationalization Crates"
 categories = ["parsing", "rendering", "encoding", "development-tools"]
 readme = "README.md"
 
+exclude = [
+    "data/**/*.txt",
+    "tools/**/*.py",
+    ".cargo/*.sh",
+    "tests/*conformance*.rs",
+]
+
 [workspace]
 
 [features]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,9 +11,11 @@ categories = ["parsing", "rendering", "encoding", "development-tools"]
 readme = "README.md"
 
 exclude = [
-    "data/**/*.txt",
-    "tools/**/*.py",
-    ".cargo/*.sh",
+    "data/**",
+    "tools/**",
+    ".cargo/**",
+    # Only ignore conformance tests;
+    # Those are the ones that rely on data files
     "tests/*conformance*.rs",
 ]
 
--- a/components/bidi/Cargo.toml
+++ b/components/bidi/Cargo.toml
@@ -9,6 +9,8 @@ keywords = ["text", "unicode", "rtl", "layout", "bidi"]
 description = "UNIC - Unicode Bidirectional Algorithm"
 readme = "README.md"
 
+exclude = [ "tests/*conformance*.rs" ]
+
 [badges]
 travis-ci = { repository = "behnam/rust-unic", branch = "master" }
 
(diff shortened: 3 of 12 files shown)


.gitignore
```diff
@@ -3,3 +3,4 @@
 *.rs.rustfmt
 target/
 Cargo.lock
+.idea/
```
behnam: I'm guessing this is the IDE keep doing this, but just a hint in case it was intentional: It's recommended to keep user-specific ignore rules under `~/.gitignore` (or whatever ignore file you set in your `~/.gitconfig`) and not under the projects. Have you tried that?

Cargo.toml
```diff
@@ -10,6 +10,13 @@ description = "UNIC - Unicode and Internationalization Crates"
 categories = ["parsing", "rendering", "encoding", "development-tools"]
 readme = "README.md"
 
+exclude = [
+    "data/**/*.txt",
+    "tools/**/*.py",
+    ".cargo/*.sh",
```
behnam: I would say lets keep it simple and ignore the whole dirs by `"data/**", "tools/**", ".cargo/**"`.

Cargo.toml
```diff
@@ -10,6 +10,13 @@ description = "UNIC - Unicode and Internationalization Crates"
...
+    "tests/*conformance*.rs",
```
behnam: Huhu. I didn't think of doing it like this. This allows keeping the `tests/unicode_version.rs` ones. Nice! (And we should document it, also.)

components/bidi/Cargo.toml
```diff
@@ -9,6 +9,8 @@ keywords = ["text", "unicode", "rtl", "layout", "bidi"]
 description = "UNIC - Unicode Bidirectional Algorithm"
 readme = "README.md"
 
+exclude = [ "tests/*conformance*.rs" ]
```
behnam: I wish we could have inheritance for workspace specs... :/

.gitignore
CAD97: Yeah, that got inserted automatically for me; I had been ignoring it manually and must have missed it this time.

Just ignoring it across my entire computer is unfortunately a non-starter; [the canonical gitignore for IDEA](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) is less restrictive than this and for my own repositories I don't mind spilling that little bit of IDE information in (basically, how to open the repository)).

I'll just keep manually ignoring; it's only 6 files and I can auto-git-add when creating the files, so it's just a mater of not accidentally adding them or letting the `.gitignore` get patched.

Cargo.toml
CAD97: 👌

components/bidi/Cargo.toml
CAD97: 👍

.gitignore
behnam: Okay, in that case, let's create a section in the gitignore file with a descriptive comment and put it there. You're making lots of contribution and there's no reason to slow it down with such a small thing.

CAD97:
Appveyor failure looks to be a server down.

.gitignore
CAD97: Actually, I just found out I can edit `../.git/info/exclude` for project specific non-committed excluding. 🎉

PR #45, opened 2017-07-20
Separate unic package from workspace root

behnam:
Making the workspace root a *virtual manifest*, only containing one direct member, `unic`, which depends on its components in return.

This would allow the workspace to contain other members next to `unic`, like the generation package, all getting builds and tests with one `--all` command.

An alternate to this would be to keep `unic` as the root (and in the top-level folder), and have the generation package a non-member directory under the root.

A third option is to make `unic` directory its own workspace, and have the generation package sitting next to the workspace.


behnam:
I think I'm going to keep this branch/PR open for a while and see if how things work in this format.

behnam:
@CAD97, what do you think about this and the alternate options mentioned above?

unic/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+license = "MIT/Apache-2.0"
+keywords = ["text", "unicode"]
+description = "UNIC - Unicode and Internationalization Crates"
+categories = ["parsing", "rendering", "encoding", "development-tools"]
+readme = "README.md"
+
+exclude = [
+    "data/**",
+    "tools/**",
+    ".cargo/**",
```
CAD97: These three exclusion lines don't mean anything anymore, since this Cargo.toml is below the point where it can see the directories `data`, `tools` and `.cargo`.

Cargo.toml
```diff
@@ -1,38 +1,5 @@
...
 [dependencies]
-unic-ucd = { path = "components/ucd/", version = "0.4.0" }
-unic-bidi = { path = "components/bidi/", version = "0.4.0" }
-unic-idna = { path = "components/idna/", version = "0.4.0" }
-unic-normal = { path = "components/normal/", version = "0.4.0" }
-
-[badges]
-travis-ci = { repository = "behnam/rust-unic", branch = "master" }
-appveyor = { repository = "behnam/rust-unic", branch = "master", service = "github" }
+unic = { path = "unic", version = "0.4.0" }
```
CAD97: If this is a virtual manifest, should it have the `[dependencies]` key? It discovers all of the packages correctly with `[workspace]`. Does this do something for just running `cargo build`?

behnam:
Cool! Yeah, I think it's a good step towards the right direction. Later, if having a *shared* workspace containing both `unic` proper and generation tools causes an issue, we can exclude the tools from the workspace, or move the workspace to the `unic/` subdir and have no manifest on the root.

I'll land after the CI passes.

PR #47, opened 2017-07-23
New table generation pipeline

CAD97:
Replaces #38.

This is a ground-up rebuilding of the table generation code, built to be extensible first.

Downloads are entirely data-driven;
`gen/config/config.toml` is a simple config file to specify what files to download and to where.

```
$ cargo run --package=unic-gen -- --help
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target\debug\unic-gen.exe --help`
unic_gen 
The UNIC Project Developers
Download data files and generate data tables for UNIC crates

USAGE:
    unic-gen [FLAGS] <components>...

FLAGS:
    -d, --download    Download the data files
    -g, --generate    Generate the data tables
    -h, --help        Prints help information
    -V, --version     Prints version information

ARGS:
    <components>...    Components to download data and generate tables for

$ cargo run --package=unic-gen -- say what
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target\debug\unic-gen.exe say what`
error: Invalid value for '<components>...': Valid components are `idna` and `ucd`, you put `say`
error: process didn't exit successfully: `target\debug\unic-gen say what` (exit code: 1)

$ cargo run --package=unic-gen -- idna ucd
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target\debug\unic-gen.exe idna ucd`
USAGE:
    unic-gen.exe [FLAGS] <components>...

Either the --download or --generate flag must be present.
For more information try --help
error: process didn't exit successfully: `target\debug\unic-gen.exe idna ucd` (exit code: 1)

$ cargo run --release --package=unic-gen -- --download --generate idna ucd
    Finished release [optimized] target(s) in 0.0 secs
     Running `target\release\unic-gen.exe --download --generate idna ucd`
Cleaning data directories...
Downloading 11 files...
>>> Loading idna Version
> unic::ucd::idna::mapping::unicode_version.rsv
>>> Loading idna IdnaMappingTable
> unic::ucd::idna::mapping::idna_mapping.rsv
>>> Loading UCD Version
>>> Loading UCD UnicodeData
> unic::ucd::age::tables::unicode_version
>>> Loading UCD DerivedAge
> unic::ucd::age::tables::age_values
> unic::ucd::bidi::tables::unicode_version
> unic::ucd::bidi::tables::bidi_class_values
> unic::ucd::category::tables::unicode_version
> unic::ucd::category::tables::general_category
> unic::ucd::core::tables::unicode_version
> unic::ucd::normal::tables::unicode_version
> unic::ucd::normal::tables::general_category_mark.rsv
> unic::ucd::normal::tables::canonical_combining_class_values.rsv
> unic::ucd::normal::tables::canonical_decomposition_mapping.rsv
> unic::ucd::normal::tables::compatibility_decomposition_mapping.rsv
```

It's also fast. It currently takes about _five seconds_ to download and generate all current tables (when compiled with optimizations).

Crates with table generation:

- [x] `unic-idna-mapping`
- [X] `unic-ucd-age`
- [X] `unic-ucd-bidi`
- [X] `unic-ucd-category`
- [X] `unic-ucd-core`
- [x] `unic-ucd-normal`

This PR does not delete the Python table generation scripts or apply the new ones. That will be done in a series of further PRs, to spread out the changes.


CAD97:
Blocking on:

- [ ] Completion of last two crates' table generation
- [ ] `char_property!` macro
  - The table generation is spitting out short names for `unic::ucd::category`. Currently, the loading code doesn't have short names in scope (or anywhere); once the `char_property!` macro is in use the names can and should be used. Thus rather than add a table into the table generation code I'm blocking on that inclusion.

CAD97:
@behnam I'm ready for a preliminary review. Other than one change (listed below) and the addition of the lookup tables emitting, I think this design is final and ready to go. (Though I'm not quite ready to merge; I'd like to be joining in without breaking the build 😛 and to have a full replacement for the python pipeline.)

In order to nicely handle multiple emit formats, my current plan is this:

- Change the current `CharMap` and `CharSet` traits into traits specifically for the bsearch table
- Add a new map for lookup tables as are currently being used

I think this paves the way most cleanly for the addition of further serialization formats without getting monolithic source files describing every single output format.

CAD97:
Because this would bother me if I didn't do it:

Proof that Rust deduplicates the slices in the generated tables (the pointers are the same):

`canonical_decomposition_mapping`: https://play.rust-lang.org/?gist=b7856ac4283a72d928171c8c8c71b111&version=stable  
Duplicates: 100  
100 of length 1

`compatibility_decomposition_mapping`: https://play.rust-lang.org/?gist=e2eb49929a0c15c1187c2d5fa403a347&version=stable  
Duplicates: 1852  
1655 of length 1  
173 of length 2  
24 of length 3

The adapter to count duplicate lengths is left as an exercise to the reader.

CAD97:
I think the best way forward is for me to revert the changes in the tables for this PR, and make a separate PR that contains the new table formats and patches the crate to use them.

I'll work on that now.

But hey, all of the generation is done!

CAD97:
I've reverted the table changes such that the tests pass and this should be merge-able.

I'll make further pull requests to start using the new tables by branching off of this branch in this coming week.

@behnam there's no more development needed on this branch, I'm fully ready for (and would appreciate) review!

gen/src/generate/idna/shared/version.rs
```diff
@@ -0,0 +1,55 @@
+// TODO: Unify this with generate::ucd::shared::version
```
behnam: The reason the data files are explicitly under `<component>/src/` is to have them *included* in packaging. If we move them out of the `<component>` dir, like `../shared/`, then the standalone packages won't build from `.crate` file.

Or maybe I misunderstood what you meant here.

gen/src/generate/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    for krate in &crates {
+        let path = tables_path(krate);
+        if path.exists() {
+            fs::remove_dir_all(&path)?;
+        }
+        fs::create_dir_all(path)?;
+    }
+
+    idna::generate_all(crates.iter().filter(|x| x.starts_with("unic-idna")))?;
```
behnam: Now that we're on it, would be great to accept the source (like `ucd` and `idna`) as a CLI argument, optionally, with default being all sources.

This would be again similar to the config matter: there could be cases that we want to download and generate files source-by-source, and having one command to download/generate all would need manual work to revert changes in git before commit.

What do you think?

gen/src/generate/ucd/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+    println!(">>> Loading UCD Version");
+    let ucd_version = shared::version::read_unicode_version()?;
+    println!(">>> Loading UCD UnicodeData");
+    let unicode_data = shared::unicode_data::read_unicode_data()?;
+
+    for krate in crates {
+        let path = super::tables_path(krate.as_ref());
+        match krate.as_ref() {
+            "unic-ucd-age" => {
+                age::generate(path, &ucd_version, &unicode_data)?;
```
behnam: Following the discussion from above...

I think there's no gain to for loop and match here, because we have a catch-all variant.

So, the for/match block can be done just as:
```
// unic-ucd-age
age::generate(path, &ucd_version, &unicode_data)?;

[...]
```

gen/src/main.rs
```diff
@@ -0,0 +1,130 @@
...
+use std::collections::HashMap;
+use std::env;
+use std::fs::File;
+use std::io::BufReader;
+use std::path::Path;
+
+use download::download;
+use generate::generate_all;
+
+use getopts::Options;
```
behnam: Should we start using <https://github.com/kbknapp/clap-rs> instead? There are many helper crates for it for various models and allowing simpler set up code/markup.

I mainly propose this, so we start using the same CLI model for all binaries, including the `src/bin/*.rs` ones. But since it's working right now, we can do this switch in a separate diff later.

gen/src/main.rs
```diff
@@ -0,0 +1,130 @@
...
+const CRATES: &'static str = "gen/config/crates.yaml";
+
+/// Take a crate name in the UNIC hierarchy and expand it into all subcrates.
+///
+/// Reads `config/crates.yaml` for the crate disassembly process.
+///
+/// Technically, a crate only needs to be in that mapping if it needs to have sources built.
+/// For consistency, all crates that need generated sources or not should be passed into the
+/// generation module, which will ignore crates it doesn't need to do anything for.
+fn expand_sub_crates<I>(list: I) -> Vec<String>
```
behnam: I think this would be unnecessary, if switch to hard-coding *target* components for each data *source*. Then, in top level, like here, we can collect those hard-coded arrays, if needed, like if we want to have a `--list` option for the CLI.

gen/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "unic-gen"
+version = "0.1.0"
+authors = ["The UNIC Project Developers"]
+homepage = "https://github.com/behnam/rust-unic/"
+repository = "https://github.com/behnam/rust-unic/"
+license = "MIT/Apache-2.0"
+description = "UNIC - Unicode Character Database - Table Generation"
```
CAD97: Whoops.

gen/config/downloads.yaml
```diff
@@ -0,0 +1,34 @@
+---
```
CAD97: It won't be difficult at all to extract the version specifiers for the different repositories. I had a vague plan to extract version numbers to a config of their own in any case, in order that you wouldn't need to always specify the correct version at the command line invocation and the repository would be self-documenting on which version of files were being used.

Along with the switch over to using TOML instead of YAML I'll reconsider the config layout.

gen/config/downloads.yaml
```diff
@@ -0,0 +1,34 @@
...
+-
+  url: "http://www.unicode.org/Public/{}/ucd/DerivedAge.txt"
```
CAD97: Since Rust doesn't have run-time format strings, it's currently handled by a simple `replace("{}", version)`. It would make sense to have it be named though, and that will just change the search string.

gen/src/download/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+///
+/// # Panics
+///
+/// Panics if the config file is missing or badly formatted, or if it fails to open a file.
+///
+/// # Errors
+///
+/// Returns a boxed error when something goes wrong during download or a write error occurs.
+pub fn download(version: &str) -> Result<(), Box<Error>> {
+    let file = File::open(Path::new(DOWNLOADS)).expect("Failed to open downloads.yaml");
```
CAD97: The reason I'm using the file open here rather than `include!()` is such that a change to the config doesn't require a recompile of the source.

Of course, in this case a change in the config will probably entail having to change the source anyway.

For the data files, it's important to read them in at runtime, because otherwise the binary can't be compiled when the data files are missing, and I like being able to download and generate with the one binary.

gen/src/generate/char_property/bsearch_set.rs
```diff
@@ -0,0 +1,90 @@
+use std::collections::BTreeSet;
+
+/// A simple range-collapsing binary search array slice.
```
CAD97: I'd argue that it makes implementation of a contains fn harder.

In the `[(low, high)]` case, you bsearch for the containing pair and if you found it, you're in the set, and if you didn't, you're out.

In the `[low, high, ]` case, you bsearch for your value, and then have to handle two cases: found it or not. In the case that you found it, if the index is even you're in, odd, you're out. In the case that you didn't find it, the index you get is the safe insertion point, so the element higher than your value, so if the index is even you're out, odd, you're in.

I think the pairs is simpler (and consistent with the value lookup).

gen/config/crates.yaml
```diff
@@ -0,0 +1,17 @@
+# For now, crates with tables we can't generate are commented out
```
CAD97: I picked YAML pretty much arbitrarily; I'm most familiar with it (maybe after JSON, which I did some work with and now have a minor personal vendetta against). I agree that TOML is probably a more consistent choice in a Rust ecosystem.

Now that I've had some time to think about it (and you've pointed out the problems with this config), it seems like the better choice is just to group by (UCD tables) / (IDNA tables) / (etc.), and have flags for generating all of the tables for a major subcrate.

There's no real need for generating for _just_ `unic-ucd-category` (other than maybe if it's just been added).

I'll do a full reconsideration of the config and command line options as I make the switch to TOML.

gen/src/generate/idna/mapping.rs
```diff
@@ -0,0 +1,143 @@
...
+        writeln!(file, "]")?;
+        Ok(())
+    }
+}
+
+impl FromStr for IdnaMapping {
+    type Err = ();
+
+    fn from_str(str: &str) -> Result<Self, Self::Err> {
+        lazy_static! {
```
CAD97: Scoping it in the function keeps the declaration close to the use site. [It's also how the regex crate suggests avoiding recompilation of the regex.](https://doc.rust-lang.org/regex/regex/index.html#example-avoid-compiling-the-same-regex-in-a-loop)

I could bring it out, but that would increase the visibility and I don't see any need to, especially since this follows the example on avoiding recompilation of regex.

gen/src/generate/idna/mapping.rs
```diff
@@ -0,0 +1,143 @@
...
+                      [[:space:]]*
+                    )?
+                  )?
+                \#"
+            ).unwrap();
+        }
+
+        /// impl note:
+        /// 8393 is the number of data lines in the Unicode version 10.0.0 IdnaMappingTable.txt
+        let mut entries = Vec::with_capacity(8393);
```
CAD97: I could always `collect` and use the length of that, but in that case there's no point because we grew the first vector in the collect.

Unfortunately, [the `captures_iter` iterator has no special information about the number of matches](https://doc.rust-lang.org/regex/regex/struct.CaptureMatches.html). We could count the number of lines and do a simple estimation that (say) 90% of lines are data lines, but that seems just as fragile.

The best answer may be to pick a nice number like 0x2200 (8704 in decimal) that gives a bit of room to grow, which also has the nice property of being one 128th of the total space of codepoints (0x110000).

gen/src/generate/idna/shared/version.rs
CAD97: Here I meant to refactor specifically `unic-gen::generate::idna::shared::version` and `unic-gen::generate::ucd::shared::version` to use the same code. They're basically the same module, just with a single string difference of which data file they're looking in. I could and probably should make them one shared file which I give the file handle or path to, rather than duplicating the file.

Of course, having the ReadMe file in the data be for the resource its downloaded and the version included in the components tables the version that its tables were generated from is what should be kept.

gen/src/generate/mod.rs
CAD97: Currently, it does do this (or a similar thing) currently.

`cargo run --package=unic-gen -- --crate unic-idna` will only generate tables for `unic-idna` and subcrates. `cargo run --package=unic-gen -- --crate unic-ucd` will only generate tables for `unic-ucd` and subcrates. `cargo run --package=unic-gen -- --unicode-version 10.0.0` will only download files as specified in the config.

Once we've moved all of the crates onto the new tables, though, I don't think this level of granularity, and just `ucd`, `idna`, `emoji`, or whatever should be fine enough. Currently it defaults to nothing done, and you `--crate unic` to specify to generate all tables; I personally would like to keep that behavior such that just running the binary will print help and it's simple to just download data without generating tables if you so choose.

gen/src/generate/ucd/mod.rs
CAD97: The reason that the loop is done is because specifying just `--crate unic-ucd-age` is possible. It does seem like that will be unnecessary once all crates are migrated to the simplified tables, though, and just handling a toggle on the first-level subcrates would be simpler.

gen/src/main.rs
CAD97: 👌 I chose `getopts` fairly arbirarily. Since I'm going to do a minor rewrite of the harness while switching to TOML, I'll go ahead and make the switch over.

gen/src/main.rs
CAD97: I'm going to need a bit more of a nudge to understand exactly what you're suggesting here.

Focusing on usage, my current idea of how the config will be laid out (if you'll excuse the vague syntax, I've not familiarized myself with the ins and outs of TOML yet):

```text
{
  idna: {
    version: "10.0.0"
    resources: [
      { url: "http://www.unicode.org/Public/idna/{version}/IdnaMappingTable.txt"
        dest: "data/idna/IdnaMappingTable.txt" }
      { url: "http://www.unicode.org/Public/idna/{version}/ReadMe.txt"
        dest: "data/idna/ReadMe.txt" }
      { url: "http://www.unicode.org/Public/idna/{version}/IdnaTest.txt"
        dest: "data/idna/test/IdnaTest.txt" }
    ]
  }
  ucd: {
    version: "10.0.0"
    resources: [
      { url: "http://www.unicode.org/Public/{}/ucd/DerivedAge.txt"
        dest: "data/ucd/DerivedAge.txt" }
    ]
  }
}
```

(and so on)

So then usage would then be

```bash
$ cargo run --crate=unic-gen -- 
# prints help

$ cargo run --crate=unic-gen -- --help
# prints help

$ cargo run --crate=unic-gen -- --download-resources
# downloads all resources

$ cargo run --crate=unic-gen -- --target idna --download-resources
# downlaods resources for idna

$ cargo run --crate=unic-gen -- --generate-tables
# generates all tables

$ cargo run --crate=unic-gen -- --target ucd --generate-tables
# generates tables for ucd
```

So the `expand_sub_crates` fn would be unnecessary by the fact that we only care about the first-level subcrates, and there is no need to expand a request for `unic-ucd` into `unic-ucd-core`, `unic-ucd-age`, etc.

CAD97: If we want to support specifying individual crate-level downloads and generation, here's my rough draft of a config and resolution path to work property:

```text
{
  ucd: {
    version: "10.0.0"
    resources: [
        { url: "http://www.unicode.org/Public/{version}/ucd/ReadMe.txt"
          dest: "data/ucd/ReadMe.txt" }
        { url: "http://www.unicode.org/Public/{version}/ucd/UnicodeData.txt"
          dest: "data/ucd/UnicodeData.txt" }
    ]
    subcrates: {
      core: []
      age: [ { url: "http://www.unicode.org/Public/{}/ucd/DerivedAge.txt"
               dest: "data/ucd/DerivedAge.txt" } ]
    }
  }
}
```

(and so on)

With this config file layout (in whatever format we go with; currently our target is TOML), the valid arguments for `--target` would be:

- Omitted (over everything)
- `ucd` (over all children of `ucd`)
- `ucd-core` (only `ucd-core` and its parents)
- `ucd-age` (only `ucd-age` and its parents)

Resolution would be the following:

- Split `--target` argument over `-`
- If length one
  - Tell that generate module to generate all subcomponents
- Else
  - Tell the generate module specified by the first part to generate subcomponents matching the latter parts

So, `--target ucd` would call (hypothetical) `generate::generate_all()` and `--target ucd-age --target ucd-core` would call `generate::generate_modules(["age", "core"])`.

Downloads would be similarly chunked based on the configuration file.

Once more, let me state that the solution in the above comment is much simpler and (I think) covers all use cases once the transition to the simplified tables has been done. It also has less moving parts and things that need to be changed when adding table generation for a new component.

Also, it's much simpler to handle the CLI arguments.

I'm perfectly willing to shoulder the transition period and the problem of only updating some tables in the git commit. I've already `git blame`d myself for all of the generated tables during this PR 😜

gen/config/downloads.yaml
behnam: > It won't be difficult at all to extract the version specifiers for the different repositories. I had a vague plan to extract version numbers to a config of their own in any case, in order that you wouldn't need to always specify the correct version at the command line invocation and the repository would be self-documenting on which version of files were being used.

Yeah, having a list of source data versions in a config files sounds the way to go!

gen/src/download/mod.rs
behnam: Sure. Makes sense.

gen/src/generate/char_property/bsearch_set.rs
behnam: Yeah, with the simple bsearch function, the pair is easier. The other one requires the index-based bsearch. Since there's not much perf/memory gain either way, let's keep it simple, as you said.

But, just to clear, the list method gets `high + 1` in the list, not `high` itself.

And, that brings back a question I faced when rewriting the Python scripts: do we want to keep these ranges closed, or half-closed? Right now they are closed, which is similar to Unicode syntax, but not much Rustic. Changing it would make reading the Rust side of the codes easier, I guess. What do you think?

gen/src/generate/idna/mapping.rs
behnam: Good point. Yeah, let's keep it here then.

gen/src/generate/idna/mapping.rs
behnam: Huhu. Yeah, I think a round number with hint like *an estimate on number of data lines [...]* would do the job. Not a big deal either way.

gen/src/generate/idna/shared/version.rs
behnam: Oh, this is a `shared` module under `unic-gen`, not `unic-ucd`. I thought it's about `src/tables/` dirs under `/unic/` components.

Sure, whatever makes it simpler.

Also, it reminded me that we *may* need API to expose the version of other data sources, as well, like IDNA version, Emoji version, etc. I'm going to file a separate issue about the implementation, but something to keep in your mind here.

gen/src/generate/mod.rs
behnam: Sounds good. One more suggestion, then, would be to make the source/target name a positional argument, since its the only main input to the cli.

By the way, I noticed that we are assuming a one-to-one mapping between data sources (like `ucd` and `idna`) and target components for the data tables (again, `ucd` and `idna`). I think that applies to majority of cases, but there could be cases that a data source name doesn't match a target component. I think we're talking about *data source names* here, right?

gen/src/generate/ucd/mod.rs
behnam: Yeah, I missed the filtering based on CLI options. If we're keeping that, it's fine then.

In the Python scripts, I intentionally didn't make it optional to generate or download only one of the tables (or anything less than all the tables) of one source, to, in some level, enforce consistency between generated files. So, for each source, all data is download at the same time (old versions `rm`ed first), and all tables generated also at the same time.

How do you feel about that approach? Do you think the time it takes to re-generate the tables could become a problem in larger components and worth loosing the enforcement?

gen/src/main.rs
behnam: Sounds great! Thanks!

gen/src/main.rs
behnam: Okay, this is exactly what I was talking about on the previous comment section. Looks like we're on the same page now.

Yeah, I agree that we should focus on download/generate actions per source, and they can be plain linear for each source, without any option to do a subset of files/tables. This would keep the CLI simpler, and maintains enforcement of updating all the data (from each source) at the same time.

So, my only feedback to the first proposal here would be:
* Let's drop the `--target`, since the value provided there is mandatory (otherwise --help happens).
* And let's not call them `target`, but `data source` or just `source`, because there's possibility of having multiple targets per source.

gen/src/generate/char_property/bsearch_set.rs
CAD97: I say keep it a closed range. The reason being border cases: since we're storing the characters bounding the range as `char` and not `u32`, if a boundary of the range is `LOW_SURROGATE - 1` or `MAX_UNICODE`, then one more for a half-open range would no longer be a valid character.

gen/src/generate/mod.rs
CAD97: I was assuming that each data source would correspond to one and only one major component.

I was specifying for the target component, but it would make just as much sense to specify by the data source. For now though, that difference is moot.

gen/src/generate/ucd/mod.rs
CAD97: Yeah, I'm switching to be a by-component specification because it's simpler. Seeing as it takes _five seconds_ (in release mode) to run currently (and not _too_ much longer in debug), I'm not too worried about reducing work right now. It's always possible to add in later.

gen/src/main.rs
CAD97: 👌 Currently this _is_ being specified by component, not data source currently. Of course, they are one to one now (as mentioned above). Other than bikeshedding one word in the help file we're in agreement though.

gen/src/generate/idna/shared/version.rs
CAD97: Seeing the plan to split versions of the different data sources makes it seems like we should keep the generating code split as well. They're different `ReadMe`s anyway, which only happen to have the same version presentation.

CAD97:
Interesting idea for cleaning up the `generate::char_property` traits a bit (removal of `_default` fn, and the associated `// FIXME`):

```rust
#[derive(Debug)]
struct DisplayWrapper<'a, T: 'a, F: 'a>(&'a T, &'a F)
where
    F: Fn(&T, &mut fmt::Formatter) -> fmt::Result;

impl<'a, T, F> fmt::Display for DisplayWrapper<'a, T, F>
where
    F: Fn(&T, &mut fmt::Formatter) -> fmt::Result,
{
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        self.1(self.0, f)
    }
}

impl<T: Eq> ToRangeBSearchMap<T> for BTreeMap<char, T> {
    fn to_range_bsearch_map<F, D>(&self, display_fn: F) -> String
    where
        F: Fn(&T, &mut fmt::Formatter) -> fmt::Result,
    {
        let mut entries = self.iter();

        let first = entries.next();
        if first.is_none() {
            return String::from("&[]");
        }

        let mut out = String::from("&[\n");
        let (mut low, mut value) = first.unwrap();
        let mut high = low;

        let append_triple = |out: &mut String, a: char, b: char, c: &T| {
            out.push_str(&format!(
                "    ('{}', '{}', {}),\n",
                a.escape_unicode(),
                b.escape_unicode(),
                DisplayWrapper(c, &display_fn),
            ));
        };

        for (char, property) in entries {
            if property != value || (*char as u32) > (*high as u32 + 1) {
                append_triple(&mut out, *low, *high, value);
                low = char;
                high = char;
                value = property;
            } else {
                assert_eq!(*char as u32, *high as u32 + 1);
                high = char;
            }
        }

        append_triple(&mut out, *low, *high, value);
        out.push_str("]");
        out
    }
}
```

What do you think @behnam? What calls were previously `to_range_bsearch_map_default()` would become `to_range_bsearch_map(fmt::Display::fmt)`. I think this cleanly models the "take a fn to format the value" while allowing consumers to pass in the standard formatting fn.

CAD97:
Just rebased from master, that should clean the history up a little.

gen/src/generate/ucd/shared/version.rs
```diff
@@ -0,0 +1,57 @@
+use std::fs::File;
+use std::io::{self, Read, Write};
+use std::path::Path;
+use std::str::FromStr;
+
+use generate::PREAMBLE;
+
+use regex::Regex;
+
+pub struct UnicodeVersion(u16, u16, u16);
```
behnam: We can put the type definition of `unic`'s `UnicodeVersion` in `unit-utils`, so we can use it here and not create a similar type with a different struct. What do you think?

CAD97: It could make sense to move `UnicodeVersion` up, but not on the case of the table generation code. I personally think that the table generation should be completely separate from the actual library. The version struct(s) will probably be the only duplication anyway, as all the other data structures in the generation code are more likely to reflect the data source than target.

PR #49, opened 2017-07-24
Move unic-ucd-utils to unic-utils-codepoints

behnam:
Create `unic-utils` component and put the Code Point-related utilities
under a new component named `unic-utils-codepoints`.

### DIFF
--- a/unic/Cargo.toml
+++ b/unic/Cargo.toml
@@ -20,10 +20,11 @@ bench_it = ["unic-bidi/bench_it"]
 serde = ["unic-bidi/serde"]
 
 [dependencies]
-unic-ucd = { path = "ucd/", version = "0.4.0" }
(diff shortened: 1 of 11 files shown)


CAD97:
Did you mean

> Create `unic-utils` component and put the Code Point-related utilities under a new component named `unic-utils-codepoints`

?

Once you merge this I'll patch #48 to crate `unic-utils-char-property` (or `unic-utils-char_property`)

behnam:
Thanks, @CAD97. Fixed.

Yeah, I though doing this first would make it easier for other `utils` crates.

I'm also going to submit a draft for char-property trait, which shows what's possible at the moment, and we can see if we want to write a separate macro, or via the `char_property` macro you've been working on, for the actual implementation.

behnam:
Good catch! Updated the doc-blocks.

PR #55, opened 2017-07-26
[travis] Drop allow_failures for nightly

behnam:
This will increase the time waiting for CI tests to pass, but we care
about the nightly test to pass and features depending on nightly to be
stable. So, better just drop `allow_failures`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,6 @@ rust:
 
 matrix:
   fast_finish: true
-  allow_failures:
-    - rust: nightly
 
 notifications:
   email:


CAD97:
Useful side effect: we can add a `rustfmt` and/or `clippy` check on the nightly CI which would fail on bad formatting or common code smells, respectively.

behnam:
`rustfmt` sounds good. `clippy` still needs some work to get there.

I think we can do these under a separate matrix config from normal nightly build, so we can easily see which part (build/tests, vs formatting) is making a problem.

PR #56, opened 2017-07-26
Apply trivial (no code updates) tables generated by unic-gen pipeline.

CAD97:

This PR only applies the trivial updates, which are just the auto-generated source comment.


CAD97:
I could slip in a change to the Python scripts to emit the new header and apply it to the other scripts as well. This would minimize time that running the Python scripts would not generate the exact same tables as are being used to only (potentially) the more major table changes that _may_ be more trouble than their worth to modify back into the Python. What do you think @behnam?

CAD97:
Closing this; We're moving piecemeal to the new tables by porting them into the python generation and can apply the rust generated files & headers once that is done.

behnam:
Thanks, @CAD97. I wanted to get back to this after my trip. Now, since it came up, let me explain a bit about what I have in mind regarding expectations from the generation tools, and we can chat more later this week.

Basically, I think it's important to maintain the whole repository *stable*, meaning that if you run the generation tools, there won't be any content changes (empty `git diff`).

One solution to maintain that for this transition period, we need to have one (bash?) script to run the python/rust-bin commands with the needed parameters to regenerate all the data. We may also need to change the tools a bit to not blindly delete all the existing data. Also, for the existing Python code, we should delete or comment-out the steps we don't need anymore, whenever that falls to the rust-bin code.

Another solution is to have one big PR, with a bunch of diffs, that switches sources (UCD and/or IDNA) to the new system. In this PR, we would kill the Python code, re-gen the tables, and update code to use the new tables.

I don't have a strong preference, at least until I look deeper into what else is left for the switch. Which one do you think is a better solution?

CAD97:
The only tables left to switch are these:

```
- unic/mapping/src/tables/idna_map.rsv
- unic/mapping/src/tables/idna_map_string.rsv
+ unic/mapping/src/tables/idna_mapping.rsv

- unic/ucd/normal/src/tables/canonical_composition_mapping_lookup.rsv
- unic/ucd/normal/src/tables/canonical_composition_mapping_values.rsv
+ unic/ucd/normal/src/tables/canonical_decomposition_mapping.rsv
- unic/ucd/normal/src/tables/
- unic/ucd/normal/src/tables/
+ unic/ucd/normal/src/tables/

**Comment by CAD97** [comment]
(I apologize if you saw the unfinished accidental post)

The only tables left to switch are these:

```
- unic/mapping/src/tables/idna_map.rsv
- unic/mapping/src/tables/idna_map_string.rsv
+ unic/mapping/src/tables/idna_mapping.rsv

- unic/ucd/normal/src/tables/canonical_composition_mapping_lookup.rsv
- unic/ucd/normal/src/tables/canonical_composition_mapping_values.rsv

+ unic/ucd/normal/src/tables/canonical_decomposition_mapping.rsv
- unic/ucd/normal/src/tables/canonical_decomposition_mapping_lookup.rsv
- unic/ucd/normal/src/tables/canonical_decomposition_mapping_values.rsv

+ unic/ucd/normal/src/tables/compatibility_decomposition_mapping.rsv
- unic/ucd/normal/src/tables/compatibility_decomposition_mapping_lookup.rsv
- unic/ucd/normal/src/tables/compatibility_decomposition_mapping_values.rsv
- unic/ucd/normal/src/tables/compatibility_decomposition_type_values.rsv
```

(And looking at that, I think I may have missed generating a canonical composition table... I'll get on that)

The differences in those changes are mostly the same: replace the two-step lookup-value lookup with a direct binary search for the value. (I'll have to do a little more digging on the composition though.)

Seeing as generating those new tables in the python would be more work than trivial (like the CCC newtype), I think it would be easier to just do one big shift to engage the rust pipeline. I could also spin together the interim patch for the python no problem, to only gen the above listed.

PR #57, opened 2017-07-26
Allow unic-ucd-normal to use unic-ucd-component

### LINKED ISSUE #53: [ucd/normal] [proposal] Allow ucd-normal to use mark information from ucd-category
`unic-ucd-normal` has its own table of symbols with a general general category of `Mark`. This duplication from `unic-ucd-category` is by design because ucd/normal does not need the entire classification that ucd/category exposes.

However, when including `unic-ucd` or just happening to want ucd/category as well, this is then duplicated information.

Hence, the proposal, which has three resolution paths:

- Add a feature (say, `ucd-category`) to `unic-ucd-normal` that, if on, does not compile the libraries own category mark table, and instead implements `is_combining_mark` as `GeneralCategory::of(character).is_mark()`
- Through benchmarks and other measurements, show that adding ucd/category as an unconditional dependency of ucd/normal is a acceptable level of bloat to avoid data duplication
- Leave it the way that it is and just test it into oblivion to make sure the implementations are in agreement (do this anyway)

Of which I suggest the first.

CAD97:
Should close #53 

This passes all current tests, but

@behnam I'm not too sure about the layout here. I haven't worked with cargo features much yet. And I don't actually know if it got tested both with and without the feature (though I think CI does that for us?)

### DIFF
--- a/unic/ucd/normal/src/gen_cat.rs
+++ b/unic/ucd/normal/src/gen_cat.rs
@@ -10,17 +10,28 @@
 // except according to those terms.
 
 
+#[cfg(not(feature = "unic-ucd-category"))]
 use std::cmp::Ordering;
-
+#[cfg(feature = "unic-ucd-category")]
+use unic_ucd_category::GeneralCategory;
 
 // General_Category = Mark
+#[cfg(not(feature = "unic-ucd-category"))]
 const GENERAL_CATEGORY_MARK: &'static [(char, char)] = include!("tables/general_category_mark.rsv");
 
 /// Return whether the given character is a combining mark (`General_Category=Mark`)
+#[cfg(not(feature = "unic-ucd-category"))]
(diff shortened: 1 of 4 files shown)


unic/ucd/normal/src/gen_cat.rs
```diff
@@ -10,17 +10,28 @@
 // except according to those terms.
 
 
+#[cfg(not(feature = "unic-ucd-category"))]
```
behnam: These should work, but also we can make it more clear by creating a local `mod` per feature-status, and put the `cfg` gate on the whole `mod`. Would be much easier to read and track relations between `use` and definitions.

CAD97: Yeah, that does seem like it would be easier.

PR #58, opened 2017-07-27
Change CanonicalCombiningClass to a newtype(u8)

### LINKED ISSUE #40: CanonicalCombiningClass should probably be a newtype
(Tuple struct, whatever)

Blocking on associated consts (we like that don't we), here's the proposed design:

```rust
struct CanonicalCombiningClass(u8);

impl CanonicalCombiningClass {
    pub const NOT_REORDERED = CanonicalCombiningClass(0);
    pub const OVERLAY = CanonicalCombiningClass(1);
    pub const NUKTA = CanonicalCombiningClass(7);
    // and so on
}

impl CanonicalCombiningClass {
    pub fn of(ch: char) -> CanonicalCombiningClass;
}

impl CanonicalCombiningClass {
    fn is_not_reordered(&self) -> bool;
}
```

Since `CanonicalCombiningClass` should is clearly a distinct concept from a `u8`, it should be a distinct type. We should use Rust's mechanisms to express that with this zero cost abstraction.

CAD97:

Closes #40 as well.

This PR applies the new table for ucd/normal/CanonicalCombiningClass, and adapts the trait to a tuple struct as discussed in #40. It also provides a `values` module with associated constants extracted from [the ccc values table](http://unicode.org/reports/tr44/#Canonical_Combining_Class_Values).

I will admit to having given names to 204, 208, 210, and 212, based on their description in the table, but I doubt anyone would object to the names I have given here.


unic/normal/src/decompose.rs
```diff
@@ -13,17 +13,18 @@
...
-fn canonical_sort(comb: &mut [(char, u8)]) {
+fn canonical_sort(comb: &mut [(char, CanonicalCombiningClass)]) {
     let len = comb.len();
     for i in 0..len {
         let mut swapped = false;
         for j in 1..len - i {
             let class_a = comb[j - 1].1;
             let class_b = comb[j].1;
-            if class_a != 0 && class_b != 0 && class_a > class_b {
+            if class_a != ccc::NotReordered && class_b != ccc::NotReordered && class_a > class_b {
```
behnam: What do you think about adding a helper method `is_reordered()`, which is `ccc != NotReordered`, and opposite of `is_not_reordered()`, to use that in these two cases?

Rational: asking people to do `!ccc.is_not_reordered()` sounds bad, and it's already used in the code, and just makes sense to have it.

unic/ucd/normal/src/canonical_combining_class.rs
```diff
@@ -12,44 +12,98 @@
...
+    /// Hiragana/Katakana voicing marks
+    pub const KanaVoicing: CCC = CCC(8);
+    /// Viramas
+    pub const Virama: CCC = CCC(9);
+    /// Marks attached at the bottom left
+    pub const AttatchedBelowLeft: CCC = CCC(200);
+    /// Marks attached directly below
+    pub const AttachedBelow: CCC = CCC(202);
+    /// Marks attached at the bottom right
+    pub const AttachedBelowRight: CCC = CCC(204);
```
behnam: I'm fine with the name, and personally like to have it. But can't think of a good way to justify the *naming* itself. Basically, we need have consistency in these situations, and I'm not sure where we are drawing the line.

IMHO:
* If we can draw a good line, we can go ahead with the name.
* If we're not sure right now, we can just leave these commented out, and open an issue to decide later.

What do you think?

unic/ucd/normal/src/canonical_combining_class.rs
```diff
@@ -12,44 +12,98 @@
...
-    /// If the *ccc* has value `Not_Reordered` (`0`).
-    fn ccc_is_not_reordered(&self) -> bool;
+#[derive(Copy, Clone, Default, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct CanonicalCombiningClass(u8);
+
+
+// TODO: Once we fully adopt 1.20 change these to associated consts on CanonicalCombiningClass
+/// Canonical Combining Classes by their name
+#[allow(non_upper_case_globals)]
+pub mod values {
```
behnam: Based on the previous discussion, should this be `long_names`, as it's aliasing Long Names for Property Values?

And, we can also add the `pub mod abbr_names`, based on <http://www.unicode.org/Public/UCD/latest/ucd/PropertyValueAliases.txt>, if you like to make it more complete.

unic/ucd/normal/src/tables/canonical_combining_class_values.rsv
```diff
@@ -1,345 +1,345 @@
-// WARNING: Auto-generated by `tools/gen_ucd_tables.py`. DO NOT EDIT MANUALLY!
+// WARNING: Auto-generated by unic-gen. DO NOT EDIT MANUALLY!
```
behnam: Could you please update the Python script to generate this, and use that instead, so we can maintain the gen scripts before landing `unic-gen`?

unic/normal/src/decompose.rs
CAD97: I say 👍 to that.

As a side note, should `is(_not)_reordered` be `#[inline]`? It's [somewhat controversial](https://stackoverflow.com/q/37639276/3019990) but this seems like the small-enough-that-we-want-it-to-be-inline-across-crates kind of thing.

unic/ucd/normal/src/canonical_combining_class.rs
CAD97: I used `values` instead of `long_names` as, as noted here, I think of this case as associated values to be used _as if_ this were a `#[repr(u8)]` enum with all values, whether named or not, valid. The enums don't use a `long_names` module, just their self namespace. In any case, I'd expect `mod long_names` to allow me to use the `Long_Name` form, rather than the `RustName` form.

As for `abbr_names`, I don't really think it's needed here. Just use `use CanonicalCombiningClass as CCC` and `CCC(10)`; it's only two more characters and might even be slightly better in matching situations, depending on how Rust treats consts in a match. (Rust requires a `Eq` implementation to use a const in a pattern context which makes me think it's a `Eq` call instead of whatever magic match over regular patterns does.)

unic/ucd/normal/src/canonical_combining_class.rs
CAD97: The guideline I used here was basically

"The Unicode document provides a description for the (unnamed) value, which follows a pattern of descriptions for other (named) values, from which the description => name transformation can be trivially derived"

I'm happy to comment them out if this isn't strong enough; we don't lose any functionality.

unic/ucd/normal/src/tables/canonical_combining_class_values.rsv
CAD97: For this module at least this is feasible and I'm happy to do it. If we want to change the preamble in the python generated tables as well see #56, where I think would be the best place to apply that specific change.

If you want to just use the Python generated tables, change all the preambles with `unic-gen`, again, just tell me, I'll close #56 and fix up this and #59 to actually use the table generated by the Python scripts.

For the broader changes introduced to the `lookup_tables` tables, I'll have to see. That's a more involved change and I still don't _quite_ grep exactly what the code generating the maps to give to `emit_lookup_tables` is doing.

(To be frank, I'd also like to avoid unnecessarily writing Python2. Unless you are a legacy project you should be using Python3, there are no dependencies still stuck on Python2 that don't have a replacement in Python3. [Python2 is also retiring in 2020.](https://pythonclock.org/) [/tangent] (sorry))

CAD97:
Potential [Level-Easy] [Area-tests] [Comp-UCD] [help wanted] after this lands: sanity check `is(_not)_reordered` against `CCC::NotReordered`, and that both aren't ever true at the same time. (If that's not too small)

Or I could divert a bit more time for that, but right now I'm trying to get everything ready to be able to land #47.

unic/ucd/normal/src/canonical_combining_class.rs
behnam: Sound reasonable, specially since it's not *the long-name*, but a local naming *based on* the long name.

Okay, let's go ahead as it is and we expand the general API guideline to fit this in.

unic/ucd/normal/src/tables/canonical_combining_class_values.rsv
behnam: Oh, I didn't mean Rust vs. Python here at all. I just meant that keeping the `master` branch stable at any time is important, and this *stable* status also refers to the table generation tools generating the same file that's committed to the branch.

In this specific case, I'm not sure if we can/should get `unic-gen` in for the `0.5` release or not. (I love to see it happen, but don't want my time constraints delay the release more.)

Anyways, it's fine. The rest of the diff is good, and the Python fix is straightforward. We can land this and fix the script to go with it.

PR #59, opened 2017-07-27
Use new tables in GeneralCategory

CAD97:

This just `use`s the variants by the short name for the purpose of loading the table. The expectation is that when the char_property macro is available to use, that will replace this block of `use`s which have no use other than to allow the table to be loaded without the generation code having to hard code in abbr=>long transformations.


unic/ucd/category/src/category.rs
```diff
@@ -80,7 +80,36 @@ pub enum GeneralCategory {
     Unassigned,
 }
 
-use self::GeneralCategory::*;
+use GeneralCategory::UppercaseLetter as Lu;
```
behnam: I thought we landed on putting these in a `pub mod abbr_names` for each component, and `use abbr_names::*` whenever needed.

Is this a different case than `BidiClass` and others?

CAD97: No, I was just doing the minimal required in order to get the new table to compile, in full expectation that the `char_property!` macro is coming soon, to replace this.

Though, since it's like three lines away from being the `abbr_names` module, I may as well do that, I now realize.

(I'll also modify the Python script to generate this changed table as well)

Issue #43, opened 2017-07-19
Expand components/ucd/tests/category_tests.rs


behnam:
We have a cross-component test in `components/ucd/tests/category_tests.rs` that checks values of the Bidi_Class property against General_Category property, based on UAX#9's [Table 4. Bidirectional Character Types](http://www.unicode.org/reports/tr9/#Table_Bidirectional_Character_Types).

Now that we have `component/ucd/category`, we can expand the test to also cover General_Category values.

CAD97:
s/utils/ucd @behnam

behnam:
Huhu. Thanks, @CAD97!

calum:
Hi, I would like to work on this. I'm new to unic but saw this in the `this-week-in-rust` update.

calum:
Hi again,

I've been looking into this issue for a couple of hours (after reading up on unicode). I'm not sure what needs to be done here. Do you want a second test which will check that all the General_Category values are correct `for cp in iter_all_chars()`? 

Or did you just want to replace the `unic_ucd::normal::is_combining_mark` with the General_Category `is_mark()` implementation? 

Thanks for any help, this project is really interesting to me so I would like to keep contributing in the future.


https://github.com/calum/rust-unic/blob/expand_catrgory_tests/unic/ucd/tests/category_tests.rs 
```rust
#[test]
fn test_bidi_nsm_against_gen_cat() {
    // Every NSM must be a GC=Mark
    for cp in iter_all_chars() {
        if BidiClass::of(cp) == BidiClass::NonspacingMark {
            assert!(GC::is_mark(&GC::of(cp))); // changed from is_combining_mark(cp);
        }
    }

    // Every GC!=Mark must not be an NSM
    for cp in iter_all_chars() {
        if !GC::is_mark(&GC::of(cp)) { // changed from !is_combining_mark(cp);
            assert_ne!(BidiClass::of(cp), BidiClass::NonspacingMark);
        }
    }
}

// This is just to help me learn what's happening
#[test]
fn test_gc_val() {
    let mut list: Vec<GC> = Vec::new();
    for cp in iter_all_chars() {
        let category = GC::of(cp);
        if !list.contains(&category) {
            list.push(category);
        }
    }
    // prints all GC values except from 'Unassigned'
    // which is obvious since we loop over all chars
    println!("{:?}", list);
}
```

CAD97:
[...]

PR #61, opened 2017-07-27
[ucd] Extend cross-component tests: GeneralCategory vs. BidiClass

calum:
This PR is my first attempt at closing #43. I have added 10 tests to the `category_tests.rs` test file.

I also corrected a typo in `category.rs` which was causing one of my tests to fail. 

Any feedback would be appreciated and I am happy to make changes.

UPDATE:
I removed my test for Bidi types OtherNeutral since from the [table of Bidi character types](http://www.unicode.org/reports/tr9/#Table_Bidirectional_Character_Types) this includes the General_Category types:

> All other characters, including OBJECT REPLACEMENT CHARACTER

Which is too general for a test - do you agree?

### DIFF
--- a/unic/ucd/tests/category_tests.rs
+++ b/unic/ucd/tests/category_tests.rs
@@ -16,10 +16,85 @@ extern crate unic_ucd;
 extern crate unic_utils;
 
 
-use unic_ucd::bidi::BidiClass;
+use unic_ucd::bidi::BidiClass as BC;
 use unic_ucd::normal::is_combining_mark;
+use unic_ucd::category::GeneralCategory as GC;
 use unic_utils::iter_all_chars;
 
+/// `normal::is_combining_mark` and `GeneralCategory::is_mark()` are expected to return
+/// the same results.
+#[test]
+fn test_gen_cat_against_normal() {
+    for cp in iter_all_chars() {
+        assert_eq!(GC::of(cp).is_mark(), is_combining_mark(cp));
+    }
+}
+
+/// <http://www.unicode.org/reports/tr9/#EN>
+#[test]
+fn test_bidi_en_against_gen_cat() {
+    for cp in iter_all_chars() {
+        if BC::of(cp) == BC::EuropeanNumber {
+            assert!(GC::of(cp).is_number());
+        }
+    }
+}
+
+/// <http://www.unicode.org/reports/tr9/#ES>
+#[test]
+fn test_bidi_es_against_gen_cat() {
+    for cp in iter_all_chars() {
+        if BC::of(cp) == BC::EuropeanSeparator {
+            assert!(
+                GC::of(cp) == GC::MathSymbol ||
+                GC::of(cp) == GC::DashPunctuation
+            );
+        }
+    }
+}
+
+/// <http://www.unicode.org/reports/tr9/#ET>
+#[test]
+fn test_bidi_et_against_gen_cat() {
+    for cp in iter_all_chars() {
+        if BC::of(cp) == BC::EuropeanTerminator {
+            assert!(
+                GC::of(cp).is_symbol()          ||
+                GC::of(cp) == GC::Unassigned    ||
+                GC::of(cp) == GC::OtherPunctuation
+            );
+        }
+    }
+}
+
+/// <http://www.unicode.org/reports/tr9/#AN>
+#[test]
+fn test_bidi_an_against_gen_cat() {
+    for cp in iter_all_chars() {
+        if BC::of(cp) == BC::ArabicNumber {
+            assert!(
+                GC::of(cp) == GC::Format            ||
+                GC::of(cp) == GC::OtherNumber       ||
+                GC::of(cp) == GC::OtherPunctuation  ||
+                GC::of(cp) == GC::DecimalNumber
+            );
+        }
+    }
+}
+
+/// <http://www.unicode.org/reports/tr9/#CS>
+#[test]
+fn test_bidi_cs_against_gen_cat() {
+    for cp in iter_all_chars() {
+        if BC::of(cp) == BC::CommonSeparator {
+            assert!(
+                GC::of(cp) == GC::OtherPunctuation  ||
+                GC::of(cp) == GC::SpaceSeparator    ||
+                GC::of(cp) == GC::MathSymbol
+            );
+        }
+    }
+}
 
 /// `Bidi_Class=NSM := General_Category in { Mn (Nonspacing_Mark), Me (Enclosing_Mark) }`
 ///
@@ -28,7 +103,7 @@ use unic_utils::iter_all_chars;
 fn test_bidi_nsm_against_gen_cat() {
     // Every NSM must be a GC=Mark
     for cp in iter_all_chars() {
-        if BidiClass::of(cp) == BidiClass::NonspacingMark {
+        if BC::of(cp) == BC::NonspacingMark {
             assert!(is_combining_mark(cp));
         }
     }
@@ -36,7 +111,54 @@ fn test_bidi_nsm_against_gen_cat() {
     // Every GC!=Mark must not be an NSM
     for cp in iter_all_chars() {
         if !is_combining_mark(cp) {
-            assert_ne!(BidiClass::of(cp), BidiClass::NonspacingMark);
+            assert_ne!(BC::of(cp), BC::NonspacingMark);
+        }
+    }
+}
+
+/// <http://www.unicode.org/reports/tr9/#BN>
+#[test]
+fn test_bidi_bn_against_gen_cat() {
+    for cp in iter_all_chars() {
+        if BC::of(cp) == BC::BoundaryNeutral {
+            assert!(GC::of(cp).is_other());
+        }
+    }
+}
+
+/// <http://www.unicode.org/reports/tr9/#B>
+#[test]
+fn test_bidi_b_against_gen_cat() {
+    for cp in iter_all_chars() {
+        if BC::of(cp) == BC::ParagraphSeparator {
+            assert!(
+                GC::of(cp) == GC::Control ||
+                GC::of(cp) == GC::ParagraphSeparator
+            );
+        }
+    }
+}
+
+/// <http://www.unicode.org/reports/tr9/#S>
+#[test]
+fn test_bidi_s_against_gen_cat() {
+    for cp in iter_all_chars() {
+        if BC::of(cp) == BC::SegmentSeparator {
+            assert!(GC::of(cp) == GC::Control);
+        }
+    }
+}
+
+/// <http://www.unicode.org/reports/tr9/#WS>
+#[test]
+fn test_bidi_ws_against_gen_cat() {
+    for cp in iter_all_chars() {
+        if BC::of(cp) == BC::WhiteSpace {
+            assert!(
+                GC::of(cp) == GC::Control           ||
+                GC::of(cp) == GC::SpaceSeparator    ||
+                GC::of(cp) == GC::LineSeparator
+            );
         }
     }
 }
--- a/unic/ucd/tests/category_tests.rs
+++ b/unic/ucd/tests/category_tests.rs
@@ -12,6 +12,9 @@
 #![cfg(test)]
 
 
+#[macro_use]
+extern crate matches;
+
 extern crate unic_ucd;
 extern crate unic_utils;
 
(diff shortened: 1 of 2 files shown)


behnam:
Thanks, @calum, for the PR! I'll do a review in the next 24hrs.

In the meanwhile, would you please make sure your `master` branch is up-to-date and your diff is rebased on top of the latest master. After rebase, you can use `git push -f` to enforce the non-fast-forward commit to github.

Also, for commit and issue/PR titles, we follow this style:
```
[<area>] Fix something
[<area>] Add another thing
```
where, the `<area>` is either a UNIC component (like `ucd` or `ucd/category`/`ucd-category`) or other things like `tools`, `travis`, etc. This is not yet mentioned in `CONTRIBUTING.md`, but we should fix soon.

unic/ucd/tests/category_tests.rs
```diff
@@ -18,8 +18,112 @@ extern crate unic_utils;
 
 use unic_ucd::bidi::BidiClass;
 use unic_ucd::normal::is_combining_mark;
+use unic_ucd::category::GeneralCategory as GC;
 use unic_utils::iter_all_chars;
 
+/// `normal::is_combining_mark` and `GeneralCategory::is_mark()` are expected to return
+/// the same results.
+#[test]
+fn test_gen_cat_against_normal() {
```
CAD97: After #57, this may be not testing the actual `ucd-normal` table for `is_combining_mark` but be calling on the implementation that calls back into `ucd-category`, because [I set `unic-ucd` to turn on the normal feature `unic-ucd-category`](https://github.com/behnam/rust-unic/pull/57/files#diff-588a4bcdb9c391c05ffe2270763cef14R22).

This test will need to either work around that or live instead in unic-normal. @behnam?

unic/ucd/tests/category_tests.rs
```diff
@@ -18,8 +18,112 @@ extern crate unic_utils;
...
+#[test]
+fn test_bidi_et_against_gen_cat() {
+    // Every BC::EuropeanTerminator must satisfy GC::is_symbol() or
+    // be GC::OtherNumber or GC::OtherPunctuation
+    for cp in iter_all_chars() {
+        if BidiClass::of(cp) == BidiClass::EuropeanTerminator {
+            assert!(
+                GC::of(cp).is_symbol()          ||
+                GC::of(cp) == GC::Unassigned    ||
+                GC::of(cp) == GC::OtherPunctuation
```
CAD97: Your test and comments don't agree here.

The doc comment says `bc=ET ⊆ ( gc=S ∪ gc=Cn ∪ gc=Po )`,  
The implementation comment says `bc=ET ⊆ ( gc=S ∪ gc=No ∪ gc=Po )`,  
The implementation says `bc=Et ⊆ ( gc=S ∪ gc=Cn ∪ gc=Po )`.

Which is it?

For what it's worth, I don't _think_ `gc=Cn` have any properties. I could be wrong though, I haven't studies the bidi data files.

calum:
My rebase keeps going wrong. So when I run ```$ git rebase upstream/master``` I get this output:

```sh
First, rewinding head to replay your work on top of it...
Applying: working on it
Applying: Added test to check that normal::is_combining_mark and GC::is_mark agree
Applying: [utils] Merge unic-utils-codepoints src into unic-utils
Using index info to reconstruct a base tree...
M	unic/ucd/age/Cargo.toml
M	unic/ucd/age/tests/age_tests.rs
M	unic/ucd/normal/Cargo.toml
M	unic/ucd/normal/tests/conformance_tests.rs
M	unic/utils/Cargo.toml
A	unic/utils/codepoints/Cargo.toml
A	unic/utils/codepoints/src/codepoints.rs
A	unic/utils/codepoints/src/lib.rs
M	unic/utils/src/lib.rs
Falling back to patching base and 3-way merge...
CONFLICT (rename/delete): unic/ucd/core/Cargo.toml deleted in [utils] Merge unic-utils-codepoints src into unic-utils and renamed in 4d5712f1e797eeb71489b140fb84d361be4d2502. Version 4d5712f1e797eeb71489b140fb84d361be4d2502 of unic/ucd/core/Cargo.toml left in tree.
CONFLICT (rename/delete): unic/idna/mapping/src/lib.rs deleted in [utils] Merge unic-utils-codepoints src into unic-utils and renamed in 4d5712f1e797eeb71489b140fb84d361be4d2502. Version 4d5712f1e797eeb71489b140fb84d361be4d2502 of unic/idna/mapping/src/lib.rs left in tree.
error: Failed to merge in the changes.
Patch failed at 0003 [utils] Merge unic-utils-codepoints src into unic-utils
The copy of the patch that failed is found in: .git/rebase-apply/patch

When you have resolved this problem, run "git rebase --continue".
If you prefer to skip this patch, run "git rebase --skip" instead.
To check out the original branch and stop rebasing, run "git rebase --abort".
```

Then when I run ```$ git status```:
```sh
rebase in progress; onto f6349c7
You are currently rebasing branch '43_extending_tests' on 'f6349c7'.
  (fix conflicts and then run "git rebase --continue")
  (use "git rebase --skip" to skip this patch)
  (use "git rebase --abort" to check out the original branch)

Unmerged paths:
  (use "git reset HEAD <file>..." to unstage)
  (use "git add <file>..." to mark resolution)

	added by us:     ../../idna/mapping/src/lib.rs
	added by us:     ../core/Cargo.toml

no changes added to commit (use "git add" and/or "git commit -a")

```

But I'm confused because I never changed those files in any of my commits? What's the best way to get around this?

@CAD97 Thanks for pointing out those typos. I've corrected them now. While writing the tests I found that quite a lot of the EuropeanTerminator were coming out as GC::Unassigned. I used the [table](http://www.unicode.org/reports/tr9/#Table_Bidirectional_Character_Types) to get a rough idea of what was expected, then wrote some ```println!()``` lines to see what was coming out,  then wrote the tests. So if a test doesn't look right to you, then I could remove it. I read that the Bidi types don't map exactly over to the Category types.

@behnam Thanks for the help with the naming. I have fixed the name of this pull request.

unic/ucd/tests/category_tests.rs
```diff
@@ -18,8 +18,112 @@ extern crate unic_utils;
...
+#[test]
+fn test_bidi_an_against_gen_cat() {
+    // Every BC::ArabicNumber must be a GC::Format, GC::OtherNumber, GC::OtherPunctuation,
+    // or GC::DecimalNumber
+    for cp in iter_all_chars() {
+        if BidiClass::of(cp) == BidiClass::ArabicNumber {
+            assert!(
+                GC::of(cp) == GC::Format            ||
+                GC::of(cp) == GC::OtherNumber       ||
+                GC::of(cp) == GC::OtherPunctuation  ||
```
behnam: Instead of `||` and `==`, you can also use `matches!()` which has a shorter syntax, and the `of()` function would get called only once.

unic/ucd/tests/category_tests.rs
```diff
@@ -18,8 +18,112 @@ extern crate unic_utils;
...
+            assert!(
+                GC::of(cp).is_symbol()          ||
+                GC::of(cp) == GC::Unassigned    ||
+                GC::of(cp) == GC::OtherPunctuation
+            );
+        }
+    }
+}
+
+/// `Bidi_Class=AN := General_Category in { Cf (Format), No (OtherNumber),
```
behnam: In general, I feel like the comments are not that helpful here and make it a bit more complicated. I think the implementation can be enough, and if there's any comment, it can be an exact copy-paste of the source, so when reading the text, we can see how it's implemented. Right now, there are two levels of interpretation, which is making me confused a bit.

In this example, using `matches` macro would give you the exact same reading in the code, as you have in the comment.

unic/ucd/tests/category_tests.rs
```diff
@@ -18,8 +18,112 @@ extern crate unic_utils;
...
+///                                         Po (OtherPunctuation), Nd (DecimalNumber) }`
+///
+/// <http://www.unicode.org/reports/tr9/#AN>
+#[test]
+fn test_bidi_an_against_gen_cat() {
+    // Every BC::ArabicNumber must be a GC::Format, GC::OtherNumber, GC::OtherPunctuation,
```
behnam: This comment is also not adding anything, and just a copy of the logic in the code.

unic/ucd/tests/category_tests.rs
```diff
@@ -18,8 +18,112 @@ extern crate unic_utils;
...
+                GC::of(cp) == GC::OtherPunctuation
+            );
+        }
+    }
+}
+
+/// `Bidi_Class=AN := General_Category in { Cf (Format), No (OtherNumber),
+///                                         Po (OtherPunctuation), Nd (DecimalNumber) }`
+///
+/// <http://www.unicode.org/reports/tr9/#AN>
```
behnam: This reference URLs are useful, though.

unic/ucd/tests/category_tests.rs
```diff
@@ -18,8 +18,112 @@ extern crate unic_utils;
 
 use unic_ucd::bidi::BidiClass;
 use unic_ucd::normal::is_combining_mark;
+use unic_ucd::category::GeneralCategory as GC;
```
behnam: Feel free to also `use ...::BidiClass as BC`, if you think makes the lines shorter and removes the clutter.

unic/ucd/tests/category_tests.rs
```diff
@@ -40,3 +144,67 @@ fn test_bidi_nsm_against_gen_cat() {
...
+fn test_bidi_bn_against_gen_cat() {
+    for cp in iter_all_chars() {
+        // Every BoundaryNeutral must satisfy GC::is_other()
+        if BidiClass::of(cp) == BidiClass::BoundaryNeutral {
+            assert!(GC::of(cp).is_other());
+        }
+
+        // Every !GC::is_other() must not be an BoundaryNeutral
+        if !GC::of(cp).is_other() {
+            assert_ne!(BidiClass::of(cp), BidiClass::BoundaryNeutral);
```
behnam: I think this assertion fails if and only if the one above fails. If so, we can drop it.

calum:
@behnam  Thanks for your feedback. I removed the unnecessary comments and some lines of code. I wanted to use the `matches!` macro, but I wasn't able to use it. Do you want me to use an external crate? I couldn't find the macro on the rust docs: (https://doc.rust-lang.org/std/#macros).

behnam:
We use the in the code link this: https://github.com/behnam/rust-unic/blob/master/unic/bidi/src/implicit.rs#L238
```
fn is_NI(class: BidiClass) -> bool {
    matches!(class, B | S | WS | ON | FSI | LRI | RLI | PDI)
}
```

First, you need to add it to the manifest (`Cargo.toml`) like this: <https://github.com/behnam/rust-unic/blob/master/unic/bidi/Cargo.toml#L26>. In this case, since it's only used in the tests, it goes under `dev-dependencies`, like this:
```
[dev-dependencies]
matches = "0.1"
```

Also, on top of the crate (the test file, in this case), you need to tell rustc to use the external resource, like this: <https://github.com/behnam/rust-unic/blob/master/unic/bidi/src/lib.rs#L72>.  Note that you need to explicitly enable macro usage:
```
#[macro_use]
extern crate matches;
```

(PS. The reason for explicit macro usage is lack of namespace system for macros. There's some discussion online about this already.)

Let me know if that resolves the issue. If not, I'll be glad to help if you provide more info.

unic/ucd/tests/category_tests.rs
```diff
@@ -16,10 +16,85 @@ extern crate unic_ucd;
...
-use unic_ucd::bidi::BidiClass;
+use unic_ucd::bidi::BidiClass as BC;
 use unic_ucd::normal::is_combining_mark;
+use unic_ucd::category::GeneralCategory as GC;
 use unic_utils::iter_all_chars;
 
+/// `normal::is_combining_mark` and `GeneralCategory::is_mark()` are expected to return
+/// the same results.
+#[test]
+fn test_gen_cat_against_normal() {
```
behnam: I have added this test to the more-specific component (`ucd-normal`) in <https://github.com/behnam/rust-unic/pull/65>, so you can now drop it from here and have this file only focus on tests for bidi-class against general-category.

calum:
@behnam I have made further changes: dropping the ucd-normal test; changing the assertions to use the `matches!` macro.

I'm happy to make any further changes if you require them.

PR #68, opened 2017-08-01
Engage unic-gen for the tables currently supported

CAD97:
If you haven't worked it out by now yet, I'm on Windows. As such, I need a double check that the shell file works as intended on native-shell machines. I'm running shell scripts when necessary through Git Bash.

The `refresh-tables.sh` calls `python2` directly because my machine uses Python3 when calling `python` (in fact this project is why I reinstalled Python2 at all). I believe modern distributions of Ubuntu's `python` symlink defaults to `python3` as well.

Potential add-on: run `refresh-tables` on CI. Would slow it down (a decent amount; it's fast when already compiled, but is (currently) compiling at `--release` level in order to be quick), but would mean we could optionally ignore the data files and/or table files in the repository, and be sure we haven't edited generated files.

This takes a conservative approach to deleting the python code which is now unused. I don't have a coverage tool to directly test and see what is or isn't now used.

(Side note, this also exposed https://github.com/CAD97/rust-unic/issues/2 which is a blocker on fully dropping the Python. Working on that.)


CAD97:
Due to the structure of the changes, #67 should still be able to merge after this without changing anything.

gen/src/generate/ucd/mod.rs
```diff
@@ -39,10 +39,12 @@ pub fn generate() -> io::Result<()> {
     fs::create_dir_all(path)?;
     core::generate(path, &ucd_version, &unicode_data)?;
 
-    let path = Path::new("unic/ucd/normal/src/tables");
-    let _ = fs::remove_dir_all(path);
-    fs::create_dir_all(path)?;
-    normal::generate(path, &ucd_version, &unicode_data)?;
+//    let path = Path::new("unic/ucd/normal/src/tables");
```
behnam: Would you please make sure the comment markers are added *indented* (not at the beginning of the line)?

tools/update_idna_data.py
```diff
@@ -1,53 +0,0 @@
-#!/usr/bin/env python
```
behnam: We don't have to delete these files in this stage, right? If so, I say lets keep them for reference, and delete when the consumer python scripts are gone.

unic/ucd/age/src/tables/age_values.rsv
```diff
@@ -1,4 +1,4 @@
-// WARNING: Auto-generated by `tools/gen_ucd_tables.py`. DO NOT EDIT MANUALLY!
+// WARNING: Auto-generated by unic-gen. DO NOT EDIT MANUALLY!
```
behnam: For someone new to UNIC, this could be a bit unclear. Can we please use this?
```
// WARNING: Auto-generated by `unic-gen` crate. DO NOT EDIT MANUALLY!
```

(I know it's not introduced in this diff, but I just noticed it since it's going in use.)

.cargo/refresh-tables.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
```
behnam: Uh, I forgot to mention that I renamed `/.cargo/` to `/etc/`, since it now has non-cargo scripts in it. We should move this to the new dir, as well.

PR #69, opened 2017-08-01
[gen] Gen normal tests

CAD97:
Closes https://github.com/CAD97/rust-unic/issues/2.

Uses unic-gen to generate `unic/normal/tests/tables` instead of the python scripts.


CAD97:
Rebased. It rebased cleanly so I'm pretty sure I did it right such that it wasn't strictly necessary.

I did have to fix one thing though -- the generated file needed the changed header.

Issue #66, opened 2017-07-31
CharProperty: API to get all possible values, if applicable


behnam:
Some properties, like `GeneralCategory` and `BidiClass`, are a flat enum, and having a type function returning all possible values of them can help with writing tests.

For example, having an iterator over GC values can reduce this test to 20 execs of the loop block, not all Unicode code points: https://github.com/behnam/rust-unic/blob/master/unic/ucd/category/tests/coverage_tests.rs

calum:
I'm interested in this one, if it's not already being worked on. 

Just to clarify: as an example, you want a function to return all chars that are `GC::MathSymbol`. Or, do you want a function to return all chars that are `GC.is_letter()`.

calum:
I've thought about it a bit more now and you actually want a function which takes a char and returns all the types it fits into? Do you want a different function for BidiClass and GeneralCategory, or one function to return all the possible values from both?

behnam:
Thanks, @calum, for the interest. The thing is that I filed this issue as a tracker for the feature, which is going to be implemented by the macros @CAD97 and I are working on. We are building tools to auto-generate most of the code that's already manually-generated, so we don't have to maintain the little things, like the feature mentioned here, manually.

There are a couple of other good starter tasks that I recommend. Let me comment there for you.

behnam:
We have the API in and there's an integration test using it, so we can track the implementation directly in the code.

PR #70, opened 2017-08-04
[ucd/category] Add all_values() static fn

behnam:
The `all_values()` static function is a new API for character
properties implemented as *flat* `enum` types in Rust, like some of
Unicode properties of type *Catalog* or *Enumeration*, to return an
exhaustive list of all possible values.

The function is useful mainly in writing integration tests, but since is
efficient in implementation, can be used in application logic, as well.

The current list is manually created and maintained. With migration to
`char_property` macros, the list will be auto-generated. (Possibly
stored in a private module and only exposed via `all_values()` API.)

See <https://github.com/behnam/rust-unic/issues/66>

### DIFF
--- a/unic/ucd/category/Cargo.toml
+++ b/unic/ucd/category/Cargo.toml
@@ -16,5 +16,4 @@ travis-ci = { repository = "behnam/rust-unic", branch = "master" }
 
 [dependencies]
 unic-ucd-core = { path = "../core/", version = "0.4.0" }
-unic-utils = { path = "../../utils/", version = "0.4.0" }
 matches = "0.1.6"
--- a/unic/ucd/category/src/category.rs
+++ b/unic/ucd/category/src/category.rs
@@ -118,9 +118,49 @@ const GENERAL_CATEGORY_TABLE: &'static [(char, char, GeneralCategory)] =
     include!("tables/general_category.rsv");
 
 impl GeneralCategory {
-    /// Find the GeneralCategory of a single char.
+    /// Find the `GeneralCategory` of a single char.
     pub fn of(ch: char) -> GeneralCategory {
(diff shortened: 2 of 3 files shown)


behnam:
@CAD97, I don't think we want to wait until rustc:1.20 becomes stable for these kinds of API additions, so, instead of the approach taken in <https://github.com/behnam/rust-unic/pull/48/commits/96c0d8173899e0f6fd1518674b1db0444c4d9a26>, I think we can go ahead with a static function for now.

We can deprecate these and use `associated_consts` when they become available, but don't think it can happen in time for unic:1.0.

What do you think?

CAD97:
@behnam what should the return type be? The simple case `&'static [$name]` doesn't seem to work: https://play.rust-lang.org/?gist=556164442d8689d67e37b043cedda9f3&version=stable

```rust
#[derive(Copy, Clone, Debug)]
enum E {
    A,
    B,
    C,
    D,
}

impl E {
    fn all_values() -> &'static [E] {
        &[E::A, E::B, E::C, E::D]
    }
}

fn main() {
    for value in E::all_values() {
        println!("{:?}", value);
    }
}
```



```text
   Compiling playground v0.0.1 (file:///playground)
error[E0597]: borrowed value does not live long enough
  --> src/main.rs:11:10
   |
11 |         &[E::A, E::B, E::C, E::D]
   |          ^^^^^^^^^^^^^^^^^^^^^^^^ does not live long enough
12 |     }
   |     - temporary value only lives until here
   |
   = note: borrowed value must be valid for the static lifetime...

error: aborting due to previous error(s)

error: Could not compile `playground`.

To learn more, run the command again with --verbose.
```

CAD97:
Yes, this looks good to me. I'll modify my `char_property` PR to use this form as well.

PR #85, opened 2017-08-06
Return a DoubleEndedIterator from iter_all_chars

CAD97:
This PR originally proposed a more drastic change. See c061708576c5989f20820a449a05e2023ed91887. It now takes the more conservative change proposed as an alternative in the original text. See the original proposal text below.

<hr/>

This is mostly to show what the iterator would look like. Having the custom iterator comes with the benefits of having a custom iterator that implements `ExactSizeIterator` and `DoubleEndedIterator`.

Note that when `impl DoubleEndedIterator` is possible, changing `iter_all_chars` to the following gets a `DoubleEndedIterator` without a custom type:

```rust
pub fn iter_all_chars() -> impl DoubleEndedIterator<Item=char> {
    (CODEPOINTS_RANGE.start..SURROGATE_RANGE.start)
        .chain(SURROGATE_RANGE.end..CODEPOINTS_RANGE.end)
        .map(|codepoint| {
            debug_assert!(char::from_u32(codepoint).is_some());
            unsafe { char::from_u32_unchecked(codepoint) }
        })
}
```

I suspect that the solution of chaining together the two scalar value ranges is going to be marginally faster than the solution in this PR which has to check every time whether or not we are about to step into the surrogate range to avoid returning invalid characters. This would need benchmarking, though, and is heavily dependent on how the `Chain` iterator moves from one iterator to the next.

Even if this PR is not taken as is, I would suggest revising `iter_all_chars` to the `impl Trait` version above in `Box<Trait>` form, as we can then provide reverse iteration at no cost.

This PR leaves `iter_all_chars` returning `Box<Iterator>` in order to not break the API, even though it would now be possible and not too problematic to return a stack value. It may be worthwhile to also, whether this proposal is accepted or not, wrap the current iterator stack into a newtype that impls `Iterator` and any other relevant iterator traits, such that the returned iterator can be stack allocated rather than boxed.

This PR also cleans up some of the comments which had gotten slightly out of date. Personally, I also find the style shown in this PR of using the `SURROGATE_RANGE` and `CODEPOINTS_RANGE` to simulate `SCALAR_VALUE_RANGE_1` and `SCALAR_VALUE_RANGE_2` conveys intent more clearly. See above for what the current chaining call looks like with these ranges.

Also, one other small negative of this PR: it quadruples the length of this file 🙃 (and removes the `#[deny(unsafe)]` from the utils crate).

### DIFF
--- a/unic/utils/src/codepoints.rs
+++ b/unic/utils/src/codepoints.rs
@@ -34,108 +34,23 @@ pub const CODEPOINTS_RANGE: Range<u32> = 0x0..(0x10_FFFF + 1);
 /// Reference: <http://unicode.org/glossary/#surrogate_code_point>
 pub const SURROGATE_RANGE: Range<u32> = 0xD800..(0xDFFF + 1);
 
-/// Check a code-point against `SURROGATE_CODEPOINTS_RANGE`.
 #[inline]
-pub fn iter_all_chars() -> Box<Iterator<Item = char>> {
-    Box::new(CharIterator::new())
-}
-
-#[derive(Copy, Clone, Debug, Ord, PartialOrd, Eq, PartialEq)]
-/// An iterator over all `char`.
-pub struct CharIterator {
-    /// The codepoint of the smallest uniterated codepoint.
-    ///
-    /// If greater than or equal to `end`, iteration is finished.
-    ///
-    /// # Safety
-    ///
-    /// Must be a valid, non-surrogate codepoint.
-    current: u32,
-
-    /// The codepoint one greater than the largest uniterated codepoint.
-    ///
-    /// If less than or equal to `current`, iteration is finished.
-    ///
-    /// # Safety
-    ///
-    /// Must be a valid, non-surrogate codepoint.
-    end: u32,
-}
-
-impl CharIterator {
-    fn new() -> CharIterator {
-        CharIterator {
-            current: CODEPOINTS_RANGE.start,
-            end: CODEPOINTS_RANGE.end,
-        }
-    }
-
-    // cannot be is_empty because usages conflict the unstable is_empty
-    fn is_finished(&self) -> bool {
-        self.current >= self.end
-    }
-}
-
-impl Default for CharIterator {
-    fn default() -> CharIterator {
-        CharIterator::new()
-    }
-}
-
-impl Iterator for CharIterator {
-    type Item = char;
-
-    fn next(&mut self) -> Option<Self::Item> {
-        if self.is_finished() {
-            return None;
-        }
-
-        let next = unsafe { char::from_u32_unchecked(self.current) };
-
-        self.current += 1;
-        if self.current == SURROGATE_RANGE.start {
-            self.current = SURROGATE_RANGE.end
-        }
-
-        Some(next)
-    }
-
-    fn size_hint(&self) -> (usize, Option<usize>) {
-        let len = self.len();
-        (len, Some(len))
-    }
-}
-
-impl DoubleEndedIterator for CharIterator {
-    fn next_back(&mut self) -> Option<Self::Item> {
-        if self.is_finished() {
-            return None;
-        }
-
-        if self.end == SURROGATE_RANGE.end {
-            self.end = SURROGATE_RANGE.start;
-        }
-        self.end -= 1;
-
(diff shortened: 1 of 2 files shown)


unic/utils/src/lib.rs
```diff
@@ -9,7 +9,7 @@
 // except according to those terms.
 
 
-#![forbid(unsafe_code, missing_docs)]
+#![forbid(missing_docs)]
```
behnam: Let's make unsafe usage more explicit. Here's how we can do ti: add
```
#![deny(unsafe)]
```
here, then on the module with unsafe need, add
```
#![allow(unsafe)]
```

unic/utils/src/codepoints.rs
```diff
@@ -29,19 +29,194 @@ use std::ops::Range;
...
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let len = self.len();
+        (len, Some(len))
+    }
+}
+
+impl DoubleEndedIterator for CharIterator {
+    fn next_back(&mut self) -> Option<Self::Item> {
+        if self.is_finished() {
```
behnam: Does this depend on underflowing of `u32`? I think that's not good practice and actually can/will panic in debug builds. (See <https://doc.rust-lang.org/std/num/struct.Wrapping.html>)

CAD97: It doesn't ever try to underflow.

If `self.end` is 0, `self.is_finished()` is true by necessity since `self.end` must be less than or equal to `self.current` since both are the same unsigned type and `self.end` is the smallest possible value.

behnam: Uh, it's the `self.end` that's moving here. Okay, looks fine then.

unic/utils/src/codepoints.rs
```diff
@@ -29,19 +29,194 @@ use std::ops::Range;
...
+    /// Must be a valid, non-surrogate codepoint.
+    current: u32,
+
+    /// The codepoint one greater than the largest uniterated codepoint.
+    ///
+    /// If less than or equal to `current`, iteration is finished.
+    ///
+    /// # Safety
+    ///
+    /// Must be a valid, non-surrogate codepoint.
```
behnam: This one is a `+1` of a valid code point, right?

CAD97: Yes, actually. `end` is one past the last uniterated codepoint. If we are to keep this I guess the correct safety requirement would be _Must be a valid, non-surrogate codepoint, or one greater than `char::MAX`_ for `end` and _Must not be a surrogate codepoint_ for `start`.

CAD97:
This was mainly exploratory, and the PR exists in order for this discussion thread to exist.

I pretty much agree that this isn't needed; I did it to see what it would look like. If I hadn't done this, I wouldn't have found the smaller diff option (listed in the first post), which honestly seems like the better option.

Switching from returning `Box<Iterator<Item=char>>` to `Box<DoubleEndedIterator<Item=char>>` would also be an API expansion, as `DoubleEndedIterator : Iterator` and is therefor only more specific.

I think the best way forward is to adopt a works-on-stable version of the fn outlined in the original post. @behnam, do you think a `Box<DoubleEndedIterator<Item=char>>` or a `struct CharIterator(..)` return would be better? I'll submit the modified PR sometime tomorrow, and close this one at that time.

behnam:
The only reason `iter_all_chars()` is returning a `Box` right now is that that's the only way to do so for trait objects. Now, whether to have a trait object, or implement a type for it, my only preference is to keep the code smaller, unless a feature is needed. Looks like we're fine with the trait object right now and we can always change it to a type later.

So, do we have any good reason to change?

CAD97:
I think that having a reversible iterator for no extra cost is useful, since we technically have one already, if we just assert that we don't need to filter out any non-character codepoints and use `map` instead of `filter_map`. This also adds tests to check that the implementation does actually work as advertised and cleans up some of the comments to be more accurate. Now that I've gone to the conservative approach rather than the all-in, I think this is worth merging.

CAD97:
(To be fully explicit, the benefit of this PR is being able to `rev`erse the iterator returned from `iter_all_chars`. To be fully honest, I don't know exactly when this would be used, but it's at zero cost.)

PR #87, opened 2017-08-08
[utils] Consolidate table bsearching

CAD97:
This attempts to consolidate all the different places that a `bsearch_range_value_table`-like fn was defined to one canonical definition in `unic-utils`.

### DIFF
--- a/unic/ucd/bidi/src/bidi_class.rs
+++ b/unic/ucd/bidi/src/bidi_class.rs
@@ -9,9 +9,9 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use std::cmp::Ordering;
 use std::fmt;
 
+use unic_utils::CharBsearchTable;
 
 /// Represents the Unicode character
 /// [*Bidi_Class*](http://www.unicode.org/reports/tr44/#Bidi_Class) property, also known as the
@@ -109,7 +109,9 @@ pub enum BidiClassCategory {
 impl BidiClass {
     /// Find the character *Bidi_Class* property value.
     pub fn of(ch: char) -> BidiClass {
-        bsearch_range_value_table(ch, BIDI_CLASS_TABLE)
+        // UCD/extracted/DerivedBidiClass.txt: "All code points not explicitly listed
+        // for Bidi_Class have the value Left_To_Right (L)."
+        *BIDI_CLASS_TABLE.binary_search_or(ch, &L)
     }
 
     /// Abbreviated name of the *Bidi_Class* property value.
@@ -211,24 +213,6 @@ impl BidiClass {
             AL | R | RLE | RLO | RLI => true,
             _ => false,
         }
-    }
-}
-
-fn bsearch_range_value_table(c: char, r: &'static [(char, char, BidiClass)]) -> BidiClass {
-    match r.binary_search_by(|&(lo, hi, _)| if lo <= c && c <= hi {
-        Ordering::Equal
-    } else if hi < c {
-        Ordering::Less
-    } else {
-        Ordering::Greater
-    }) {
-        Ok(idx) => {
-            let (_, _, cat) = r[idx];
-            cat
-        }
-        // UCD/extracted/DerivedBidiClass.txt: "All code points not explicitly listed
-        // for Bidi_Class have the value Left_To_Right (L)."
-        Err(_) => L,
     }
 }
 
--- a/unic/ucd/bidi/src/bidi_class.rs
+++ b/unic/ucd/bidi/src/bidi_class.rs
@@ -11,7 +11,7 @@
 
 use std::fmt;
 
-use unic_utils::CharBsearchTable;
+use unic_utils::CharDataTable;
 
 /// Represents the Unicode character
 /// [*Bidi_Class*](http://www.unicode.org/reports/tr44/#Bidi_Class) property, also known as the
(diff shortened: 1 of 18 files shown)


unic/utils/src/tables.rs
```diff
@@ -0,0 +1,56 @@
...
+impl CharBsearchTable<()> for &'static [(char, char)] {
+    fn find(&self, needle: char) -> Option<&()> {
+        self.binary_search_by(|&(low, high)| if needle < low {
+            cmp::Ordering::Less
+        } else if needle > high {
+            cmp::Ordering::Greater
+        } else {
+            cmp::Ordering::Equal
+        }).map(|_| {
+                const UNIT: &() = &();
```
behnam: Why not just return `()` here, so we don't have to have a `const` variable defined and referenced to?

IMHO, returning `Option<_>` from all these endpoints is enough consistency, and there's no need to try to keep reference-ness of the type also be the same. IIRC, there are similar APIs in the std lib, and reference is only returned if there's actual value and it makes sense to do so.

What do you think?

CAD97: The reason that it's `&()` here is because the definition of `fn find` returns `Option<&V>`.

I just found the way to lay out the lifetimes to get it to work returning `Option<V>`, testing for push now.

unic/utils/src/tables.rs
```diff
@@ -0,0 +1,56 @@
...
+
+use std::cmp;
+
+/// bananas
+pub trait CharBsearchTable<V> {
+    /// bananas
+    fn find(&self, needle: char) -> Option<V>;
+
+    /// bananas
+    fn binary_search_or(&self, needle: char, default: V) -> V {
```
behnam: Looks like this can be `find_or()`, and the next `find_or_else`.

unic/utils/src/tables.rs
```diff
@@ -0,0 +1,56 @@
...
+
+impl CharBsearchTable<()> for &'static [(char, char)] {
+    fn find(&self, needle: char) -> Option<()> {
+        self.binary_search_by(|&(low, high)| if low > needle {
+            cmp::Ordering::Greater
+        } else if high < needle {
+            cmp::Ordering::Less
+        } else {
+            cmp::Ordering::Equal
+        }).map(|_| ())
```
behnam: 🎉

CAD97:
@behnam Just one proposal before I do the docs:

Would it make sense to drop `binary_search_or`/`binary_search_or_else`? They just replace `find.unwrap_or`/`find.unwrap_or_else` respectively. Even if we want to keep them for ease of use, should they be renamed to `find_or`/`find_or_else`? The algorithm used to `find` the element is a binary search, but it also might make sense to rename the trait from `CharBsearchTable` to `CharDataTable`, and maybe even make some impls for other table formats that we decide to use.

There are three possible changes to add to this PR before merging listed below, what are your opinions?

- Rename `binary_search_or` and `binary_search_or_else` to `find_or` and `find_else`
  - This closer matches the actual API of the trait
- Remove `binary_search_or` and `binary_search_or_else`
  - As can be seen in their implementation, all the necessary expressiveness already comes from returning an `Option` from `find`
- Rename `CharBsearchTable` to `CharDataTable`
  - This better captures the API of the trait which doesn't really require a binary search, just a data extraction.
  - Then this trait would also be expandable to using things like PHF for data tables with no change in usage

unic/utils/src/tables.rs
CAD97: 👍 You ninja'd me asking about that

CAD97:
Going with both renames, but keep the `_or(_else)` for ease of use. Now just to doc it properly and then we're good for merge.

behnam:
Huhu. We were commenting at the same time.

About method names, yeah, let's have `find_or` and `find_or_else`, since it's common pattern. (You see similar *completeness* in the std lib API, as well.)

About the trait name, I think it's a good idea to make it more generic, but I say let's land this and work on that part in a separate diff. Specially, since I have another idea to add to that...

Right now, the nature of the tables are hidden in the components, because only the *structure* of the table is exposed. I think we should switch to newtype patterns, having `BoolBsearchTable` instead of `&'static [(char, char)]`, `RangedBearchTable` (or something like that) instead of `[(char, char, V)]`, and `SingleBsearchTable` instead of `&'a [(char, V)]`. Then, the trait can be very generic and have implementation for each of these, only exposing high-level methods like `find*()` and hiding the table details behind it.

What do you think?

CAD97:
Sounds good to me! Though if we're going to newtype it, we should probably add a `CharRange(char, char)` as well rather than continue using a `(char, char)`... but that's for the next bit of work.

behnam:
With `CharRange`, and your passion to implement iterators (😉), we can implement an iterator for `CharRange` (and also `IntoIterator` for it), so we can have `for cp in CharRange::all()`, instead of `for cp in iter_all_chars()`!  This would make a very useful abstraction that's missing from core libraries, and we can provide publicly here. (I think there's a similar third-party crate, but not double-ended.)

unic/ucd/bidi/src/bidi_class.rs
```diff
@@ -109,7 +109,9 @@ pub enum BidiClassCategory {
 impl BidiClass {
     /// Find the character *Bidi_Class* property value.
     pub fn of(ch: char) -> BidiClass {
-        bsearch_range_value_table(ch, BIDI_CLASS_TABLE)
+        // UCD/extracted/DerivedBidiClass.txt: "All code points not explicitly listed
+        // for Bidi_Class have the value Left_To_Right (L)."
+        *BIDI_CLASS_TABLE.find_or(ch, &L)
```
behnam: Let's land this, but also one comment for later:

I believe this is the job for `Default::default()` and we may want to actually switch to that in all cases. What do you think?

CAD97: If we always pack the [complex default values](http://unicode.org/reports/tr44/#Default_Values_Table) into the tables, then definitely yes. The only reason `find_or_else` exists is an expectation that it might be used for one of these complex defaults.

behnam: Right. Let's keep it as it is for now, then, and we can clean up after adding a couple more of properties.

PR #89, opened 2017-08-08
[char_property] Implement *CharProperty traits

behnam:
Character Properties are of different kinds and shapes, and as UNIC
components grow, we need a better way to be able to categorize them by
their shape, and a way to make sure we have consistent, noncolliding
API for them.

This is the first step into building a CharProperty taxonomy, with as
little as possibly needed to provide the assurances desired.

We hope that the implementation can be improved over time with new
features added to the language. There's already some proposals in this
front. See these discussions for more details:

* [Traits as contract, without changes to call-sites](https://users.rust-lang.org/t/traits-as-contract-without-changes-to-call-sites/11938/11)

* [RFC: delegation of implementation](https://github.com/rust-lang/rfcs#1406)

### DIFF
--- a/unic/utils/src/char_property.rs
+++ b/unic/utils/src/char_property.rs
@@ -20,41 +20,35 @@ use std::fmt::{Debug, Display};
 use std::hash::Hash;
 
 
-// == Completeness and Accessors ==
-
-/// Marker for character property types.
-pub trait CharProperty<T> {}
-
-// Make char property type T also a parameter to this trait, otherwise there will be "conflicting
-// implementation" error when marking traits below as `CharProperty`.
-impl<T> CharProperty<T> for T {}
-
-
-/// A Character Property defined on all characters.
-///
-/// Examples: *Age*, *Name*, *General_Category*, *Bidi_Class*
-pub trait CompleteCharProperty
-where
-    Self: Copy + Debug + Default + Display + Eq + Hash,
-{
-    /// Find the character property value.
-    fn of(ch: char) -> Self;
-}
-
+// == Property Domain ==
 
 /// A Character Property defined for some characters.
 ///
 /// Examples: *Decomposition_Type*, *Numeric_Type*
-pub trait PartialCharProperty
-where
-    Self: Copy + Debug + Display + Eq + Hash,
-{
+pub trait PartialCharProperty: Copy + Debug + Display + Eq + Hash {
     /// Find the character property value, or None.
     fn of(ch: char) -> Option<Self>;
 }
 
 
-// == Enumerated/Catalog Property Type ==
+/// A Character Property defined on all characters.
+///
+/// Examples: *Age*, *Name*, *General_Category*, *Bidi_Class*
+pub trait CompleteCharProperty: PartialCharProperty + Default {
+    /// Find the character property value.
+    fn of(ch: char) -> Self;
+}
+
+impl<T: CompleteCharProperty> PartialCharProperty for T {
+    fn of(ch: char) -> Option<Self> {
+        Some(<Self as CompleteCharProperty>::of(ch))
+    }
+}
+
+
+// == Property Range ==
+
+// === Enumerated/Catalog Types ===
 
 /// A Character Property with enumerated values.
 ///
@@ -63,16 +57,13 @@ where
 /// Usage Note: If the property is of type *Catalog*, it's recommended to (in some way) mark the
 /// type as *non-exhaustive*, so that adding new variants to the `enum` type won't result in API
 /// breakage.
-pub trait EnumeratedCharProperty<T: CharProperty<T>>: Sized
-where
-    Self: CharProperty<T>,
-{
+pub trait EnumeratedCharProperty: Sized + PartialCharProperty {
     /// Exhaustive list of all property values.
     fn all_values() -> &'static [Self];
 }
 
 
-// == Numeric Property Type ==
+// === Numeric Types ===
 
 /// Marker for numeric types accepted by `NumericCharProperty`.
 pub trait NumericCharPropertyValue {}
@@ -83,10 +74,8 @@ impl NumericCharPropertyValue for u8 {}
 /// A Character Property with numeric values.
 ///
 /// Examples: *Numeric_Value*, *Canonical_Combining_Class*
-pub trait NumericCharProperty<T: CharProperty<T>, Value: NumericCharPropertyValue>
-where
-    Self: CharProperty<T>,
-{
+pub trait NumericCharProperty<Value: NumericCharPropertyValue>
+    : PartialCharProperty {
     /// Get numeric value for character property value
     fn number(&self) -> Value;
 }
--- a/unic/utils/src/char_property.rs
+++ b/unic/utils/src/char_property.rs
@@ -9,38 +9,71 @@
 // except according to those terms.
 
 
-//! TBD.
+//! Taxonomy and Contracts for Character Property types.
+//!
+//! See also the list of types of character properties defined in the UCD:
+//! <http://unicode.org/reports/tr44/#About_Property_Table>, in [Unicode® Standard Annex #44 —
+//! Unicode Character Database](http://unicode.org/reports/tr44/#About_Property_Table)
 
 
 use std::fmt::{Debug, Display};
 use std::hash::Hash;
 
 
-/// TBD.
+/// A Character Property defined on all characters.
+///
+/// Examples: *Age*, *Name*, *General_Category*, *Bidi_Class*
 pub trait CharProperty
 where
     Self: Copy + Debug + Default + Display + Eq + Hash,
 {
-    /// TBD
+    /// Find the character property value.
     fn of(ch: char) -> Self;
 }
 
 
-/// TBD.
+/// A Character Property defined for some characters.
+///
+/// Examples: *Decomposition_Type*, *Numeric_Type*
 pub trait OptionCharProperty
 where
     Self: Copy + Debug + Display + Eq + Hash,
 {
-    /// TBD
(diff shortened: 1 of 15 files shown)


CAD97:
The first link gives me a "Oops! That page doesn’t exist or is private." I think you have a trailing `>` which got included in the URL. https://users.rust-lang.org/t/traits-as-contract-without-changes-to-call-sites/11938/11

The second link follows a couple redirects and then 404s. Since the URL starts with behnam/rust-unic, I suspect github's intra-github linking autocomplete got the better of you. rust-lang/rfcs/pull/1406

behnam:
Oops. Fixed the links!

behnam:
Okay, I think I have covered all the basics of all the UCD properties we already have and now this PR is ready for review and land.

Here's the summary of implementations:
```rust
$ g g -e 'impl \w*CharProperty' unic/ucd/
unic/ucd/age/src/age.rs:impl CharProperty for Age {
unic/ucd/bidi/src/bidi_class.rs:impl CharProperty for BidiClass {
unic/ucd/bidi/src/bidi_class.rs:impl EnumeratedCharProperty for BidiClass {
unic/ucd/category/src/category.rs:impl CharProperty for GeneralCategory {
unic/ucd/category/src/category.rs:impl EnumeratedCharProperty for GeneralCategory {
unic/ucd/normal/src/canonical_combining_class.rs:impl CharProperty for CanonicalCombiningClass {
unic/ucd/normal/src/canonical_combining_class.rs:impl NumericCharProperty<u8> for CanonicalCombiningClass {
unic/ucd/normal/src/decomposition_type.rs:impl OptionCharProperty for DecompositionType {
unic/ucd/normal/src/decomposition_type.rs:impl EnumeratedCharProperty for DecompositionType {
```

What do you think, @CAD97?

unic/utils/src/char_property.rs
```diff
@@ -0,0 +1,79 @@
...
+
+
+use std::fmt::{Debug, Display};
+use std::hash::Hash;
+
+
+/// A Character Property defined on all characters.
+///
+/// Examples: *Age*, *Name*, *General_Category*, *Bidi_Class*
+pub trait CharProperty
```
CAD97: This should probably extend from `OptionCharProperty`, as do `Eq : PartialEq` and `Ord : PartialOrd`.

I would then do a blanket impl from `CharProperty` to `OptionCharProperty`:

```rust
impl<P: CharProperty> OpionCharProperty for P {
    fn of(ch: char) -> Option<Self> {
        Some(<Self as CharProperty>::of(ch))
    }
}
```

[Really rough playground proof of concept](https://play.rust-lang.org/?gist=25da653487d1713a01d1aa0f83773c6d&version=stable)

unic/utils/src/char_property.rs
```diff
@@ -0,0 +1,79 @@
...
+
+
+/// A Character Property with enumerated values.
+///
+/// This is similar to types *Catalog* and *Enumeration*, as defined in UAX#44.
+///
+/// Usage Note: If the property is of type *Catalog* (as defined by Unicode), it's recommended to
+/// (in some way) mark the type as *non-exhaustive*, so that adding new variants to the `enum` type
+/// won't result in API breakage.
+pub trait EnumeratedCharProperty
```
CAD97: This should probably extend from `(Option)CharProperty`, unless it would ever make sense to have a `EnumeratedCharProperty` which is not also a `(Option)CharProperty`.

unic/utils/src/char_property.rs
```diff
@@ -0,0 +1,79 @@
...
+/// Marker for numeric types accepted by `NumericCharProperty`.
+pub trait NumericCharPropertyValue {}
+
+impl NumericCharPropertyValue for u8 {}
+
+
+/// A Character Property with numeric values.
+///
+/// Examples: *Numeric_Value*, *Canonical_Combining_Class*
+pub trait NumericCharProperty<Value>
```
CAD97: This should probably extend from `(Option)CharProperty`, unless it would ever make sense to have a `NumericCharProperty` which is not also a `(Option)CharProperty`.

unic/utils/src/char_property.rs
```diff
@@ -0,0 +1,79 @@
...
+where
+    Self: Copy + Debug + Default + Display + Eq + Hash,
+    Value: NumericCharPropertyValue,
```
CAD97: It's a matter of taste, but I would put this trait bound inside the `<>`.

Reasoning: it's simple, and is very important for the definition of the trait.

unic/ucd/normal/src/gen_cat.rs
```diff
@@ -13,12 +13,10 @@
...
     use std::cmp::Ordering;
 
-    const GENERAL_CATEGORY_MARK: &'static [(char, char)] =
-        include!("tables/general_category_mark.rsv");
-
     /// Return whether the given character is a combining mark (`General_Category=Mark`)
     pub fn is_combining_mark(c: char) -> bool {
-        GENERAL_CATEGORY_MARK.find(c).is_some()
+        const TABLE: &'static [(char, char)] = include!("tables/general_category_mark.rsv");
+        TABLE.find(c).is_some()
```
CAD97: Wait how is this working when you aren't `use`ing `unic_utils::CharDataTable` anywhere in the file?

CAD97: I got it working with `use unic_utils::CharDataTable;` under `mod mark` and `CharDataTable::<()>::find(&TABLE, c).is_some()`.

Unfortunately `&[(char, char)]` matches `&[(char, V)]` as well, thus the reason I want to get around to adding the `CharRange` structure.

Once all of `unic-gen` is engaged and I can make the transition to actually using it across the board, I'll work on that.

unic/utils/src/char_property.rs
behnam: The thing about `Eq`/`Ord` and their `Partial` equivalents is that the function signature stays the same, and it's only the semantics that different.

Here, we have a different signature for `of`, which we can pull of to work somehow, but I don't think it's good practice.

Have you seen anything more like this case on std/3rd-party libs?

unic/utils/src/char_property.rs
behnam: Yeah, I think I can introduce a marker here to ensure more enforcement.

unic/ucd/normal/src/gen_cat.rs
behnam: Okay, I'm splitting the clean up diff, plus a fix for this, into another PR. Will update this PR with only the char-props matters.

unic/utils/src/char_property.rs
CAD97: I guess the only real example is the (ongoing?) discussion of an `impl<T: From> TryFrom for T`.

I'd be fine without this, but this would also simplify the marker for `Enumerated/NumericCharProperty` to just be `OptionCharProperty` since `CharProperty` implies `CharProperty`.

I don't really have an argument for it, but it just seems like "infallible transformation from char" is a subset of "fallible transformation from char". (Thus, the same use case as `Result<_, !>`?)

unic/utils/src/char_property.rs
```diff
@@ -0,0 +1,92 @@
...
+
+
+// == Completeness and Accessors ==
+
+/// Marker for character property types.
+pub trait CharProperty<T> {}
+
+// Make char property type T also a parameter to this trait, otherwise there will be "conflicting
+// implementation" error when marking traits below as `CharProperty`.
+impl<T> CharProperty<T> for T {}
```
behnam: Actually, this makes the marker useless, since every T now is a `CharProperty`! :( Need to figure out another solution here...

CAD97: Seems like what you want is `CompleteCharProperty` and `OptionCharProperty` to be mutually exclusive...

Making `OptionCharProperty` a parent of `CompleteCharProperty` and using `OptionCharProperty` would solve the problem (and in any case I suspect `of` will be intrinsic and used as such), but it seems difficult to use this pattern like this.

behnam: Yeah, can't figure out a way to tell compiler that these two traits are mutually exclusive.

What do you mean by "I suspect `of()` will be intrinsic and used as such"?

unic/utils/src/char_property.rs
```diff
@@ -0,0 +1,81 @@
...
+pub trait PartialCharProperty: Copy + Debug + Display + Eq + Hash {
+    /// Find the character property value, or None.
+    fn of(ch: char) -> Option<Self>;
+}
+
+
+/// A Character Property defined on all characters.
+///
+/// Examples: *Age*, *Name*, *General_Category*, *Bidi_Class*
+pub trait CompleteCharProperty: PartialCharProperty + Default {
```
behnam: This works very well. Thanks for the idea, @CAD97!

The only issue is the error if none of these are implemented and a Property Range type is used, in which case rustc claims that `CompleteCharProperty` is required to be implemented, which is invalid, and only `PartialCharProperty` is required. I suppose I should file an issue for this upstream.

unic/utils/src/char_property.rs
CAD97: I expect to see 

```rust
enum Property {}
impl Property {
  fn of()...
}
impl CharProperty for Property {
  fn of() = (Self as Property).of()...
}
```

if that works such that you can just do `Property::of` without knowing about `CompleteCharProperty`.

But  ¯\\\_(ツ)_/¯ I'm just assuming shapes of unwritten things right now

behnam:
Posted about the side issues notice here on the forum: https://users.rust-lang.org/t/unexpected-behaviors-of-trait-bounds/12286

PR #92, opened 2017-08-09
[ucd/normal] Use the right `.find()` method

behnam:


behnam:
I reverted the changes related to moving the const tables inside the `of()` functions, because I remembered that we want to add to the CharProperty API methods to retrieve values for `CharRange`, which is more efficient than single lookups.

PR #94, opened 2017-08-09
Ucd normal new tables

CAD97:
Currently fails in idna.

I can't see what I messed up, and ucd/normal itself passes its own tests. Any ideas, @behnam?

Git history got a little messy because I grabbed charprop but reverted it again; I'll squash everything before the master merge into one commit before merge to clean up that mess.


CAD97:
Closing while I work on some history-rewriting things

PR #96, opened 2017-08-09
[char_property] Manually enforce CompleteCharProperty inherited traits

behnam:
See this bug report for why we need to do this at the moment: https://github.com/rust-lang/rust/issues/43777


behnam:
bors: r+

PR #98, opened 2017-08-10
Use new tables for [ucd/normal]

CAD97:
Uses the simplified table layout generated by `unic-gen`.


CAD97:
@behnam The tests are failing in punycode, but I didn't touch it. All of ucd/normal's tests still pass, and I can't find what I did to break punycode. Could you take a look?

CAD97:
uh, how is this both mergeable and not? Did I break AppVeyor with my history rewriting?

unic/ucd/normal/src/composition.rs
```diff
@@ -12,65 +12,35 @@
...
 }
 
 // == Compatibility Decomposition (KD) ==
-const COMPATIBILITY_DECOMPOSITION_LOOKUP: &'static [(char, Slice)] =
-    include!("tables/compatibility_decomposition_mapping_lookup.rsv");
-const COMPATIBILITY_DECOMPOSITION_VALUES: &'static [char] =
-    include!("tables/compatibility_decomposition_mapping_values.rsv");
+use DecompositionType::*;
+use DecompositionType::Nobreak as NoBreak;
+pub(crate) const COMPATIBILITY_DECOMPOSITION_MAPPING: &'static [(
```
behnam: `pub(crate)` would break it for `rustc:1.17.0`, right? I think it was stabilized in `1.19.0`.

unic/ucd/normal/src/composition.rs
```diff
@@ -12,65 +12,35 @@
...
+    CANONICAL_DECOMPOSITION_MAPPING.find(c).map(|it| *it)
 }
 
 // == Compatibility Decomposition (KD) ==
-const COMPATIBILITY_DECOMPOSITION_LOOKUP: &'static [(char, Slice)] =
-    include!("tables/compatibility_decomposition_mapping_lookup.rsv");
-const COMPATIBILITY_DECOMPOSITION_VALUES: &'static [char] =
-    include!("tables/compatibility_decomposition_mapping_values.rsv");
+use DecompositionType::*;
+use DecompositionType::Nobreak as NoBreak;
```
behnam: nit: I would just fix this issue in the generation side, and don't special-case it here.

Looking at all the long_names for the value,
 looks safe to just *to_lower* and *capitalize* all DT value names. (http://www.unicode.org/Public/UCD/latest/ucd/PropertyValueAliases.txt)

unic/ucd/normal/src/composition.rs
CAD97: Sounds right to me. (yes.)

What's the global visibility here? This _really should not_ leak outside the crate but it needs to be shared between `composition.rs` and `decomposition.rs`.

behnam: One solution would be to put all these tables in one module, as `pub`, and only consume that module internally.

Or, don't `pub mod composition` in the first place and only `pub use` whatever we really want to expose. (In other words, make `composition` an internal module.)

CAD97: Honestly, it seems like the `de/composition` mods should be private. I'll make that change. (Wait, they are... I was being overly defensive.)

CAD97:
(Rebased on top of the `char_property` merge.)

gen/src/generate/mod.rs
```diff
@@ -17,3 +17,11 @@ pub mod ucd;
 /// The standard auto-generated source disclaimer.
 const PREAMBLE: &'static str = "// WARNING: Auto-generated by the `unic-gen` crate.\n\
                                 // WARNING: DO NOT EDIT MANUALLY!";
+
+fn capitalize(str: &str) -> String {
```
behnam: This reminds me that we may also want the same logic for `FromStr`, since, by UAX#44 definition, all property/value names are case-insensitive. Or, we can implement a `find()` function or something like that, that does more extensive search for value, so that we keep `FromStr` performant. What do you think?

CAD97: [std::ascii::AsciiExt::eq_ignore_ascii_case](https://doc.rust-lang.org/std/ascii/trait.AsciiExt.html#tymethod.eq_ignore_ascii_case)

Since property names are guaranteed to be ASCII only, this should work... (with some tweaks)

```rust
match str {
    $( stringify!($variant) => Ok($name::$variant), )*
    $( stringify!($abbr) => Ok($name::$abbr_variant), )*
    $( stringify!($long) => Ok($name::$long_variant), )*
    $( str if str.eq_ignore_ascii_case( stringify!($variant) ) => Ok($name::$variant), )*
    $( str if str.eq_ignore_ascii_case( stringify!($abbr) ) => Ok($name::$abbr_variant), )*
    $( str if str.eq_ignore_ascii_case( stringify!($long) ) => Ok($name::$long_variant), )*
    _ => Err(()),
}
```

Rust guarantees that when two match arms could both match, the first is taken, so it'll do the quicker matches check before the more expensive eq_ignore_ascii_case. I'll go try it out and send the PR.

behnam:
Okay, try `git amend` and `git push` again. I changed some GH repo settings which should fix AppVeyor issues.

CAD97:
I found the problem!

The string `"אבּג"` is failing `unic::idna::process::validate` on this branch with validation `V6`.

Adding the following debug code:

```rust
eprintln!("-----");
eprintln!("{}", label);
eprintln!(
    "{:#?}",
    label.chars()
        .map(|c| c.escape_unicode().collect::<String>())
        .collect::<Vec<_>>()
);
eprintln!("{:#?}", label.chars().map(Mapping::of).collect::<Vec<_>>());
```

to the entry of the fn showed this difference between master and this branch:

```text
# master
-----
אבּג
[
    "\\u{5d0}",
    "\\u{5d1}",
    "\\u{5bc}",
    "\\u{5d2}"
]
[
    Valid,
    Valid,
    Valid,
    Valid
]

# ucd-normal-new-tables
-----
אבּג
[
    "\\u{5d0}",
    "\\u{fb31}",
    "\\u{5d2}"
]
[
    Valid,
    Mapped(
        StringTableSlice {
            byte_start: 8020,
            byte_len: 4
        }
    ),
    Valid
]
```

The validated string used to be

```text
  ‎05D0  HEBREW LETTER ALEF
  ‎05D1  HEBREW LETTER BET
  ‎05BC  HEBREW POINT DAGESH OR MAPIQ
  ‎05D2  HEBREW LETTER GIMEL
```

Now the validated string is

```text
  ‎05D0  HEBREW LETTER ALEF
  ‎FB31  HEBREW LETTER BET WITH DAGESH
  ‎05D2  HEBREW LETTER GIMEL
```

The form in the string literal has not changed.

I've found the root of the error; the string being processed is different. @behnam, I'm going to have to defer to you here because I don't know enough of how the idna crate is laid out to trace this much further. Though it does just look like the `nfc` form of the string in question changed with the new tables...

CAD97:
(Rebased on top of master since I had some changes added to this branch which got in by other commits)

behnam:
Cool! No worries. Let me take a look at it later today or tomorrow, then. Right now I'm trying to figure out how none of the non-idna tests are failing if something's going wrong in a non-idna component.

CAD97:
@behnam That's easy: it's because the NFC form of the string changed and no tests covered the particular character(s) which changed.

CAD97:
I found it! (Struggling to not blow up with excitement)

<http://www.unicode.org/Public/10.0.0/ucd/CompositionExclusions.txt>  
<http://www.unicode.org/reports/tr15/#Primary_Exclusion_List_Table>

`HEBREW LETTER BET WITH DAGESH` is in the list of characters in the Composition Exclusion Table

Why it was working before I've yet to figure out but I found it!

behnam:
Nice! Well, that specific part of the Exclusions was in the Python code already. Maybe we missed it in the Rust gen?

tools/gen_ucd_tables.py
```diff
@@ -1,347 +0,0 @@
...
-
-@memoize
-def get_normal_props():
-    interestingprops = ["Full_Composition_Exclusion"]
-
-    props = {}
-    re1 = re.compile("^ *([0-9A-F]+) *; *(\w+)")
-    re2 = re.compile("^ *([0-9A-F]+)\.\.([0-9A-F]+) *; *(\w+)")
-
-    for line in fileinput.input(data_path("DerivedNormalizationProps.txt")):
```
behnam: Here's what's missing from the code.

Btw, we should put some unit tests for this part of the code to make sure this doesn't happen again!

CAD97:
I must have missed it. In fact the rust generation code never touches `DerivedNormalizationProps.txt`.

tools/gen_ucd_tables.py
CAD97: The actual libraries (and tests) sort of are the tests for the generation code. I'm definitely adding some tests into `ucd-normal` to make sure the excluded compositions don't compose though!

CAD97:
I'm probably going to refactor out `DerivedNormalizationProps` into a shared block of data like `UnicodeData`, but for now it just lives in the ucd-normal generation.

All the tests pass locally, we should be good to (review one more time and) merge @behnam

gen/src/generate/ucd/normal.rs
```diff
@@ -147,10 +191,11 @@ impl CanonicalCompositionData {
             }),
         )
     }
-}
 
-impl<'a, 'b> From<&'a CanonicalDecompositionData<'b>> for CanonicalCompositionData {
```
behnam: What happened here? I'm not sure why this `From` was needed, but also don't understand why to have a non-`From` impl of `from`. Are we de-generalizing the impl? (Too bad we don't have a test for this expection...)

CAD97: What happened is I needed to take two things, the decomposition and the exclusion.

I suppose we could have a `impl From<(Decomposition, Exclusions)>` (or just call it `new`).

behnam:
bors: r+

PR #100, opened 2017-08-10
[utils] Make char_property FromStr impl ascii case insensitive

CAD97:
[`std::ascii::AsciiExt::eq_ignore_asci_case`](https://doc.rust-lang.org/std/ascii/trait.AsciiExt.html#tymethod.eq_ignore_ascii_case)

We still check for simple equality first, so simple equality should still be fast.


behnam:
bors: r+

PR #101, opened 2017-08-10
[char_property] Create module dir and add submodules

behnam:
Put the contract traits and macro all next to each other under one
module dir.

Also update some documentations.


behnam:
bors: r? @CAD97

CAD97:
bors: r+

CAD97:
(I don't think I have permission in bors (whether I should or not is a different question), but I wanted to try)

behnam:
> (I don't think I have permission in bors (whether I should or not is a different question), but I wanted to try)

*Now* you have! :)

behnam:
bors: r+

behnam:
bors: r+

notriddle:
@behnam Why do you have both [bors] and [bors-voyager] enabled on the same repo?

[bors]: https://github.com/apps/bors
[bors-voyager]: https://github.com/apps/bors-voyager

behnam:
That was a mistake, coming out of my explorations. Dropped @bors-voyager access already.

Btw, thanks for the great work with borg-ng, @notriddle! It's helping us a lot already!

PR #103, opened 2017-08-10
WIP: [ucd] Implement Name char property

behnam:
It's a WIP. One problem with it right now is that it increases the compile time by dozens of seconds, because of the optimization being done with the `&[&'static str]` table. Because of that, we may be better off with a manually implemented mapping, meaning that table generation would take some time, but component compilation stays fast.


CAD97:
I don't know if this will actually work to lower the amount of work rustc is doing, but you might be able to just manually deduplicate the `&'static str`:

```rust
const NAMES: &'static [&'static [&'static str]] = {
    const LATIN: &'static str = "LATIN";
    const CAPITAL: &'static str = "CAPITAL";
    const LETTER: &'static str = "LETTER";
    &[
        &[LATIN, CAPITAL, LETTER, "A"],
        &[LATIN, CAPITAL, LETTER, "B"],
        &[LATIN, CAPITAL, LETTER, "C"],
    ]
};
```
[playground link](https://play.rust-lang.org/?gist=1df8a36a5d83129c69507f78bf28d46d&version=stable)

Also, something of note:

> When a string value not enclosed in <angle brackets> occurs in this field, it specifies the character's Name property value, which matches exactly the name published in the code charts. The Name property value for most ideographic characters and for Hangul syllables is derived instead by various rules. See _Section 4.8, Name_ in [Unicode] for a full specification of those rules. Strings enclosed in <angle brackets> in this field either provide label information used in the name derivation rules, or—in the case of characters which have a null string as their Name property value, such as control characters—provide other information about their code point type.
[[source](http://unicode.org/reports/tr44/#Name)]

The full name algorithm will need to either handle these algorithmically or precalculate the names.

HOWEVER!

There has been much prior work done in compressing Unicode character names, one of which methods can very likely be recycled here, rather than trying to roll our own:

- Rust: @huonw's [`unicode_names`](https://crates.io/crates/unicode_names) (Last updated to Unicode 7.0)
- Python: [`unicodedata`](https://docs.python.org/library/unicodedata.html) (Last updated to Unicode 9.0.0)
- Java: [ICU4J UCharacterName](https://ssl.icu-project.org/repos/icu/trunk/icu4j/main/classes/core/src/com/ibm/icu/impl/UCharacterName.java) (if you can make any sense of it)

I think the best choice is to analyze `unicode_names` and use the compression technique @huonw chose three years ago.

See also a couple resources I found attempting to find the ICU character name compression scheme:

- [Compact Encoding of Names](http://site.icu-project.org/design/compactenc/names)

behnam:
Right. There are some open questions here. I want to put this out as an experiment with the rustc's lowering feature.

So, there are two concepts here: one is *Name*, and another is the unofficial *Label*. *Name* is defined by UCD, and is an optional char property: not defined for all chars, like unassigned chars, some CJK, etc. UCD Names are unique and stable.

Labels, on the other hand, are usually composed: if Name is available, it's equal to the Name. if char has no Name, then it's composed from the hints published in `UnicodeData.txt` and the character codepoint (if uniqueness is desired).

I suppose the same component (`unic-ucd-name`) can provide both of these.

About implementation, yeah, we should look at existing methods and choose one. I haven't spent time on that yet.

CAD97:
It could be interesting to see how a simple `&[(char, &[&str])]` binary search array compares to a "more intelligent" method; most of the deduplication will be accomplished by the method you've used here of separating over the whitespace.

Having had a bit of time to think about it, this may actually match (or nearly match) the compression of other methods that only are a mapping of name to character. We couldn't do the reverse, but I don't think that's in scope of what we're trying to provide (right now). I'm going to experiment a little and see if I can avoid the compilation slowdown.

CAD97:
I have a version which is completing `cargo build` in ~10s on my machine (debug mode):

<https://github.com/behnam/rust-unic/compare/master...CAD97:character-name-experiment>

It's not an actual component because it was merely a test to see if I could reduce the work required to deduplicate the slices.

behnam:
Cool! ~10s is much much better. Would you like to take over and merge this PR with that diff and send a new PR, so we can have a working version and work on the API, while considering how to optimize the tables?

Btw, about the two table files, I guess you can put both in an `rsv` file, in a block, like this:
```
// WARNING: Auto-generated by the `unic-gen` crate.
// WARNING: DO NOT EDIT MANUALLY!
{
const _A: &str = "-A";
const _CHAL: &str = "-CHAL";
const _CHAR: &str = "-CHAR";
const _DZUD: &str = "-DZUD";
const _KHYIL: &str = "-KHYIL";
const _KHYUD: &str = "-KHYUD";
// [...]

&[
    ('\u{20}', &[SPACE]),
    ('\u{21}', &[EXCLAMATION, MARK]),
    ('\u{22}', &[QUOTATION, MARK]),
    ('\u{23}', &[NUMBER, SIGN]),
    ('\u{24}', &[DOLLAR, SIGN]),
    ('\u{25}', &[PERCENT, SIGN]),
    // [...]
]
}
```

This should give the exact same compile time, I guess.

CAD97:
This doesn't seem to work, though I will attempt it one more time and see if I inadvertently changed something:

<https://github.com/CAD97/rust-unic/blob/b545763cf63d59cec35d4f4032291168aa7b46d1/unic/ucd/name/src/tables/name_values.rsv> by <https://github.com/CAD97/rust-unic/blob/b545763cf63d59cec35d4f4032291168aa7b46d1/gen/src/generate/ucd/name.rs>

I'll test that and see how it behaves, then submit a new PR.

CAD97:
I have run the compilation with the declarations put into an inline block with the slice and compilation speed tanked again. This seems like a `rustc` bug. I can make a small self-contained example, but where would I report this behavior?

behnam:
Hum... interesting! Yeah, could be a bug in rustc. When we have a small repro diff for before/after, we can file it on <https://github.com/rust-lang/rust/issues/>

behnam:
Being worked on in https://github.com/behnam/rust-unic/pull/125

PR #104, opened 2017-08-10
Fix ucd-normal without ucd-category

CAD97:
`.find()` is ambiguous here. `[(char, char)]` (intended) and `for<V> [(char, V)]` both match.


behnam:
bors: r+

PR #105, opened 2017-08-10
[ucd/normal] Add use-ucd-category feature

behnam:
Basically, the dev dep record has been auto-enabling the optional dev
record for all test builds, resulting in the error to not surface.

Fix for the error surfacing by this change is
<https://github.com/behnam/rust-unic/pull/104>.

The original issue, which is (IMHO) unexpected Cargo behavior is filed
here: <https://github.com/rust-lang/cargo/issues/4395>


behnam:
bors: try

CAD97:
bors: r+

behnam:
Actually, this got me thinking now: having the feature (*using UCD category instead of local table file*) enabled whenever the package is available in the build could actually be what we want here. The downside is that won't be able to have a test for it, but the benefit is not having a separate `use-ucd-category` everywhere there's a deep/shallow dependency to `unic-ucd-normal`.

I'm going to send another PR for that now.

PR #106, opened 2017-08-11
[ucd/normal] Drop use-ucd-category feature

behnam:
Revert of <https://github.com/behnam/rust-unic/pull/105>

Also, add a `![cfg()]` to the integration test, to only perform it when
it matters. Also, this way, we can see if the feature is auto-enabled or
not. (Which is not!)


behnam:
Thanks, @CAD97!

bors: r+

behnam:
bors: r+

PR #108, opened 2017-08-11
[ucd/normal] Rename to decomposition_type_conformance_tests.rs

behnam:
If we call a file `conformance_tests.rs`, then it should cover
conformance of all data in the component. Otherwise, the test file name
should specify the module name it's testing.


behnam:
bors: r+

PR #109, opened 2017-08-11
[idna][normal] Add use-ucd-category feature

behnam:
`ucd-normal` has the option to use `ucd-category`. Therefore, it makes
sense to enable the feature whenever we know `ucd-category` is
available, which include `unic` proper.

In a larger scope, I'm not sure if we want to go down this path, and may
actually worth it if we just drop the optional feature and always do
that. But, in another hand, it's a good excercise for optional
cross-component dependencies and we can decide later depending on what
other kinds of options we add.


behnam:
bors: r+

PR #111, opened 2017-08-12
CharRange and CharIter

CAD97:
Expanded documentation and tests are TODO -- working on those presently.

We can land the patch to use the char range in a different PR.

### DIFF
--- a/unic/char/range/Cargo.toml
+++ b/unic/char/range/Cargo.toml
@@ -1,0 +1,27 @@
+[package]
+name = "unic-char-range"
+version = "0.5.0"
+authors = ["The UNIC Project Developers"]
+repository = "https://github.com/behnam/rust-unic/"
+license = "MIT/Apache-2.0"
+keywords = ["text", "unicode", "iteration"]
+description = "UNIC - CharRange"
+categories = ["text-processing", "iteration"]
+readme = "README.md"
+
+# No tests/benches that depends on /data/
+exclude = []
+
+[features]
+default = []
+
+# Unstable features
+unstable = [ "associated-consts", "fused", "trusted-len" ]
+associated-consts = []
+fused = []
+trusted-len = []
+
+[dependencies]
+
+[badges]
+travis-ci = { repository = "behnam/rust-unic", branch = "master" }
--- a/unic/char/range/src/iter.rs
+++ b/unic/char/range/src/iter.rs
@@ -1,5 +1,10 @@
 use std::{char, iter};
-use {CharRange, SURROGATE_RANGE};
+use CharRange;
+
+/// Range of Surrogate Code Points.
+///
+/// Reference: <http://unicode.org/glossary/#surrogate_code_point>
+const SURROGATE_RANGE: ::std::ops::Range<u32> = 0xD800..0xE000;
 
 /// An iterator over `char`.
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
--- a/unic/char/range/src/iter.rs
+++ b/unic/char/range/src/iter.rs
@@ -47,6 +47,7 @@ impl CharIter {
 impl Iterator for CharIter {
     type Item = char;
 
+    #[inline]
     #[allow(unsafe_code)]
     fn next(&mut self) -> Option<char> {
         if self.is_finished() {
--- a/unic/char/range/src/iter.rs
+++ b/unic/char/range/src/iter.rs
@@ -45,7 +45,9 @@ impl Iterator for CharIter {
 
     #[allow(unsafe_code)]
     fn next(&mut self) -> Option<char> {
-        if self.is_finished() { return None }
+        if self.is_finished() {
+            return None;
+        }
 
         let char = unsafe { char::from_u32_unchecked(self.low) };
         self.low += 1;
@@ -67,7 +69,9 @@ impl Iterator for CharIter {
 impl iter::DoubleEndedIterator for CharIter {
     #[allow(unsafe_code)]
     fn next_back(&mut self) -> Option<char> {
-        if self.is_finished() { return None }
+        if self.is_finished() {
+            return None;
+        }
 
         self.high -= 1;
         let char = unsafe { char::from_u32_unchecked(self.high) };
@@ -83,7 +87,9 @@ impl iter::DoubleEndedIterator for CharIter {
 
 impl iter::ExactSizeIterator for CharIter {
     fn len(&self) -> usize {
-        if self.is_finished() { return 0 }
+        if self.is_finished() {
+            return 0;
+        }
         let naive_len = self.high as usize - self.low as usize;
         if self.low <= SURROGATE_RANGE.start && SURROGATE_RANGE.end <= self.high {
             naive_len - SURROGATE_RANGE.len()
--- a/unic/char/range/src/iter.rs
+++ b/unic/char/range/src/iter.rs
@@ -24,11 +24,10 @@ pub struct CharIter {
 }
 
 impl<'a> From<&'a CharRange> for CharIter {
-    #[inline]
     fn from(range: &CharRange) -> CharIter {
         CharIter {
-            low: range.start as u32,
-            high: range.end as u32 + 1,
+            low: range.first() as u32,
+            high: range.last() as u32 + 1,
         }
     }
 }
--- a/unic/char/range/src/iter.rs
+++ b/unic/char/range/src/iter.rs
@@ -1,0 +1,101 @@
+use std::{char, iter};
+use {CharRange, SURROGATE_RANGE};
+
+/// An iterator over `char`.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct CharIter {
+    /// The codepoint of the smallest uniterated codepoint.
+    ///
+    /// If greater than or equal to `high`, iteration is finished.
+    ///
+    /// # Safety
+    ///
+    /// Must be a valid, non-surrogate codepoint while iteration is unfinished.
+    low: u32,
+
+    /// The codepoint one greater than the largest uniterated codepoint.
+    ///
+    /// If less than or equal to `low`, iteration is finished.
+    ///
+    /// # Safety
+    ///
+    /// Must be one greater than a valid, non surrogate codepoint while iteration is unfinished.
+    high: u32,
+}
+
+impl<'a> From<&'a CharRange> for CharIter {
+    #[inline]
+    fn from(range: &CharRange) -> CharIter {
+        CharIter {
+            low: range.start as u32,
+            high: range.end as u32 + 1,
+        }
+    }
+}
+
+impl CharIter {
+    #[inline]
+    fn is_finished(&self) -> bool {
+        self.low >= self.high
+    }
+}
+
+impl Iterator for CharIter {
+    type Item = char;
+
+    #[allow(unsafe_code)]
+    fn next(&mut self) -> Option<char> {
+        if self.is_finished() { return None }
+
+        let char = unsafe { char::from_u32_unchecked(self.low) };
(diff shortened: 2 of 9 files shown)


unic/char/range/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "unic-char-range"
+version = "0.5.0"
+authors = ["The UNIC Project Developers"]
+repository = "https://github.com/behnam/rust-unic/"
+license = "MIT/Apache-2.0"
+keywords = ["text", "unicode", "iteration"]
+description = "UNIC - CharRange"
+categories = ["text-processing", "iteration"]
```
CAD97: Haven't actually checked that this is valid yet.

CAD97:
Benchmark results from `char-iter`:

```text
test benches::count          ... bench:   1,000,194 ns/iter (+/- 34,045)
test benches::count_baseline ... bench:     937,333 ns/iter (+/- 66,945)
```

Benchmark results from `unic-char-range`:

```text
test count          ... bench:   2,855,694 ns/iter (+/- 174,285)
test count_baseline ... bench:     941,054 ns/iter (+/- 59,795)
```

... I've got a bit of work to do don't I ...

CAD97:
I can not for the life of me figure out why `char-iter` is twice as fast as I am.

I'm going to try re-reimplementing this but closer to `char-iter` and see if I can't get the better performance.

CAD97:
I figured out where I was losing time. In my experiments I think I found a cleaner for our uses design though... and it involves fewer moving parts (I think)

unic/char/range/Cargo.toml
behnam: There's no "iteration" category. Maybe we don't need more than one. Here's the list: https://crates.io/categories/

FYI, I've got "internationalization" (and "localization") added to the list, but it hasn't made it to the live version yet. When that happens, almost all UNIC components get both `text-processing` and `internationalization`.

unic/char/range/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "unic-char-range"
+version = "0.5.0"
+authors = ["The UNIC Project Developers"]
+repository = "https://github.com/behnam/rust-unic/"
+license = "MIT/Apache-2.0"
+keywords = ["text", "unicode", "iteration"]
+description = "UNIC - CharRange"
```
behnam: Would be better to use more informative language in description, as it's the only text that shows up on crates.io lists and item page, and allows better matching in search.

To follow the pattern from other UNIC components, it can be something like this:
```
description = "UNIC - Unicode Characters - Character Range and Iteration"
```

unic/char/range/src/iter.rs
```diff
@@ -0,0 +1,112 @@
...
+    type Item = char;
+
+    #[inline]
+    #[allow(unsafe_code)]
+    fn next(&mut self) -> Option<char> {
+        if self.is_finished() {
+            return None;
+        }
+
+        let char = unsafe { char::from_u32_unchecked(self.low) };
```
behnam: I didn't know it's possible to call a variable `char`! That's cool! But also uncommon practice, but I don't know why! Anyways, may be safer to just call it `ch` or `chr`.

unic/char/range/src/iter.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+impl iter::ExactSizeIterator for CharIter {
+    fn len(&self) -> usize {
+        if self.is_finished() {
+            return 0;
+        }
+        let naive_len = self.high as usize - self.low as usize;
+        if self.low <= SURROGATE_RANGE.start && SURROGATE_RANGE.end <= self.high {
+            naive_len - SURROGATE_RANGE.len()
```
behnam: Maybe you can optimize here by putting this number in a `const` variable, but could be unnecessary. Not sure.

unic/char/range/src/iter.rs
```diff
@@ -0,0 +1,112 @@
...
+        if self.is_finished() {
+            return None;
+        }
+
+        self.high -= 1;
+        let char = unsafe { char::from_u32_unchecked(self.high) };
+
+        // ensure `high` is never one greater than a surrogate code point
+        if self.high == SURROGATE_RANGE.end {
+            self.high = SURROGATE_RANGE.start;
```
behnam: This implementation with the new variants is even more clean and slim! 👍

unic/char/range/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+//! }
+//!
+//! for character in CharRange::all() {
+//!     // character is every valid char from lowest codepoint to highest
+//! }
+//! # }
+//! ```
+//!
+#![forbid(bad_style, missing_debug_implementations, unconditional_recursion)]
+#![deny(missing_docs, unsafe_code, unused, future_incompatible)]
```
behnam: Why `unused, future_incompatible` are in `deny()` but not `forbid()`? Doesn't look like we `allow()` them anywhere in this crate.

unic/char/range/src/range.rs
```diff
@@ -0,0 +1,153 @@
...
+use {step, CharIter};
+
+// This is a closed range to match the pattern used in text processing rather than in Rust's Range:
+// regex `['a'-'z']` and unicode ranges `U+0000 - U+10FFFF` are inclusive on both ends.
+// Additionally, this allows us to easily store ranges which border invalid characters.
+/// A range of unicode code points.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub struct CharRange {
+    /// The lower bound of the range (inclusive).
+    first: char,
```
behnam: Why not call them `start` and `end` as in `Range*` structs?

unic/char/range/src/range.rs
```diff
@@ -0,0 +1,153 @@
...
+        last: '\u{10FFFF}',
+    };
+}
+
+/// Common collection-like fn
+impl CharRange {
+    /// Create an iterator over this range of characters.
+    #[inline]
+    pub fn iter(&self) -> CharIter {
+        CharIter::from(self)
```
behnam: `self.into()` should work, too! 😉

unic/char/range/src/range.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    /// Does this range include a character?
+    #[inline]
+    pub fn contains(&self, char: char) -> bool {
```
behnam: Same about the `char` variable name.

unic/char/range/src/step.rs
```diff
@@ -0,0 +1,40 @@
+use std::char;
```
behnam: Since we have a module here called `char`, I think it would be easier to read the code if we stick with `std::char` instead of using it as `char`. What do you think?

unic/char/range/src/step.rs
```diff
@@ -0,0 +1,40 @@
...
+
+const BEFORE_SURROGATE: char = '\u{D7FF}';
+const AFTER_SURROGATE: char = '\u{E000}';
+
+/// Step a `char` toward `char::MAX`.
+///
+/// If the given `char` is already `char::MAX`, it is returned unchanged.
+#[allow(unsafe_code)]
+pub fn forward(c: char) -> char {
```
behnam: Using the same variable name here would make is easier to read.

unic/char/range/src/step.rs
```diff
@@ -0,0 +1,40 @@
...
+    if c == char::MAX {
```
behnam: So, these checks result in an `open_range()` to silently create a `CharRange` with `first` > `last`. For example, if we go with `open_range('\u{0x20}', '\u{0x21}'), then `first == U+21` and `last = U+20`. Does this violate any invariants, or we don't care?

(It doesn't look like causing any issues right now, but may be with expansions of the API.)

unic/char/range/tests/iter_tests.rs
```diff
@@ -0,0 +1,63 @@
...
+        assert_eq!(Some(custom_char), simple.next());
+        assert_eq!(custom.next_back(), simple.next_back());
+    }
+    assert_eq!(None, simple.next());
+}
+
+#[test]
+fn iter_fused() {
+    let mut iter = CharRange::all().iter();
+    let mut fused = all_chars().into_iter().fuse();
```
behnam: I think this needs to be gated behind the `fused` feature. Surprised that it's passing the CI!

behnam:
Btw, also take a look at https://doc.rust-lang.org/std/collections/enum.Bound.html and its usage in https://doc.rust-lang.org/std/collections/range/trait.RangeArgument.html .

CAD97:
Whoops, replaced it while you were reviewing... Much was shared though so it's not lost effort.

unic/char/range/src/iter.rs
CAD97: You'll like #112 then 😄

unic/char/range/src/iter.rs
CAD97: 👍 it's a bad habit. It's not something you'd really expect to work, but it does (I think because Rust does type and variable name lookup differently). Surprisingly though, my IDE highlights it right, which is part of why I do it even though I probably shouldn't.

unic/char/range/Cargo.toml
CAD97: 👍

unic/char/range/tests/iter_tests.rs
CAD97: [The fn to fuse an iterator](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fuse) is stable. It's merely the trait to declare yourself as fused that is unstable.

unic/char/range/src/lib.rs
CAD97: d24057a for `future_incompatible`. (It doesn't work on 1.17 for some reason.) `unused`, just because I was `allow`ing it temporarily while building the library up.


Basically, I picked `deny` for things that could plausibly be temporarily allowed.

unic/char/range/src/iter.rs
CAD97: Unfortunately doing so and still calling `len()` would require `const fn` to be stable. I can see if precomputing it manually saves any non-negligible amount of time though!

CAD97:
As for the macro, the syntax was a curveball idea. I'm sticking with just the standard `..` and `..=` offerings.

unic/char/range/src/iter.rs
CAD97: Turns out it takes `2 ns (+/- 0 ns)` either way.

unic/char/range/src/step.rs
CAD97: If the problem is `std::char` and the primitive type `char` conflicting, that's an intended overlap. The `char` module contains things like `char::from_u32` and `char::MAX` which should act like they are on the primitive `char`.

If you're saying `unic-char`, that's not referenced anywhere from my code. The only time there is a `char` module there is from inside the main `unic` crate.

unic/char/range/src/lib.rs
behnam: I see. I guess it's fine for non-major flags, and we should probably just rely on clippy (in a future version) to hint on all `forbid`able `deny`s.

unic/char/range/src/step.rs
behnam: Oh, I thought we have a `char.rs` here next to `step.rs`. You're right, there's no conflict and it's fine. #aftermidnightreview

unic/char/range/src/step.rs
behnam: Cool! Yeah, as long as the behavior is intended and works similar to std lib, it's fine.

I think it worth adding a hint about it in the constructors, though.

Issue #91, opened 2017-08-09
Implement CharRange type


CAD97:
[assign: @CAD97]

Replace usages of `(char, char)` as an inclusive range with a custom [Range](https://doc.rust-lang.org/std/ops/struct.Range.html) struct adapted for use on characters.

behnam:
Filed https://github.com/behnam/rust-unic/issues/93 as a follow up to this, for `CharProperty` API accepting `CharRange` as input.

CAD97:
I found [`char-iter`](https://github.com/huonw/char-iter) on crates.io after a little digging, but that's only an iterator and not an actual range object. A cursory search found only `better-range` that actually provided a range over characters, and that crate doesn't have documentation or source links on crates.io.

(Also, both crates are two years old, but then again character iteration isn't going to change.)

In conclusion, adding a `CharRange` type will not duplicate work already done in the ecosystem.

Because a character range is such a self-contained interest (and I'm currently working on a project that would also enjoy having one), I'm considering making it its own crate. What do you think @behnam?

behnam:
Thanks for the research, @CAD97. Very helpful.

Yeah, now thinking about it again, looks like things like `char_property` and `char_range` can be standalone crates, and not just parts of `unic-utils`, so they can be easily everywhere.

That said, I like the idea of keeping all these things under one repository, because it allows faster development and one-step reviews, instead of going back and forth between repos. The main problem in this area is the fact that we don't have CI/devops tools to test cross-repo PRs before merging. Hopefully this will change in a near future, but for now, it's too much cost to maintain small repos. This is why there are many multi-package repositories (like https://github.com/BurntSushi/ripgrep/), which host many small packages internally, so they can be developed with all the cross-package tests.

I think eventually we can split out these kinds of packages into their own repo. But for the current stage, which is fast development and not maintaining a stable API yet, I think we better keep them all in.

About organization in UNIC, we can actually put these components under `/unic/char/range/` and `/unic/char/property/`, and name them `unic-char-range` and `unic-char-property`. This will keep names short, but still show which repository they come from, and keep them easy to find.

What do you think?

CAD97:
That seems like a good path forward to me!

behnam:
I think we can generally improve `Range<char>` in rustc, so we can drop this class later. But, for now, we need the functionality and this model is the best abstraction, and easier to replace with built-in libs later.

Some ideas for now...

[...]

PR #112, opened 2017-08-12
[char/range] Add CharRange and CharIter

CAD97:
The first half of adressing #91.
Closes #111, this manner of attack is better than it.

This PR only has one type, `CharRange`. It is effectively `std::ops::RangeInclusive` translated to characters. The matter of construction is handled by both half-open and closed constructors offered and a macro to allow for `'a'..='z'` syntax.

### DIFF
--- a/unic/char/range/src/range.rs
+++ b/unic/char/range/src/range.rs
@@ -1,10 +1,26 @@
+use std::char;
+use std::ops::Range;
 use step;
-use {AFTER_SURROGATE, BEFORE_SURROGATE};
+
+const SURROGATE_RANGE: Range<u32> = 0xD800..0xE000;
 
 /// A range of unicode code points.
 #[derive(Debug)]
 pub struct CharRange {
+    /// The lowest uniterated character.
+    ///
+    /// Iteration is finished if this is higher than `high`.
+    ///
+    /// # Safety
+    ///
+    /// This is not guaranteed to always be a valid character. Check before using!
+    /// Note that `high` _is_ guaranteed to be a valid character,
+    /// so this will always be a valid character when iteration is not yet finished.
     low: char,
+
+    /// The highest uniterated character.
+    ///
+    /// Iteration is finished if this is lower than `low`.
     high: char,
 }
 
@@ -19,24 +35,59 @@ impl CharRange {
     }
 }
 
+impl CharRange {
+    #[inline]
+    #[allow(unsafe_code)]
+    // It is always safe to step `self.low` forward because
+    // it will only be used when less than `self.high`.
+    fn step_forward(&mut self) {
+        self.low = unsafe { step::forward(self.low) }
+    }
+
+    #[inline]
+    #[allow(unsafe_code)]
+    // When stepping `self.high` backward would cause underflow,
+    // step `self.low` forward instead. It will have the same effect --
+    // consuming the last element from the iterator and ending iteration.
+    fn step_backward(&mut self) {
+        if self.high == '\0' {
+            self.step_forward();
+        } else {
+            self.high = unsafe { step::backward(self.high) }
+        }
+    }
+}
+
 impl Iterator for CharRange {
     type Item = char;
 
     #[inline]
-    #[allow(unsafe_code)]
     fn next(&mut self) -> Option<char> {
         if self.low > self.high {
             return None;
         }
 
         let char = self.low;
-        self.low = unsafe { step::forward(self.low) };
+        self.step_forward();
         Some(char)
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
         let len = self.len();
         (len, Some(len))
+    }
+}
+
+impl DoubleEndedIterator for CharRange {
+    #[inline]
+    fn next_back(&mut self) -> Option<Self::Item> {
+        if self.low > self.high {
+            return None;
+        }
+
+        let char = self.high;
+        self.step_backward();
+        Some(char)
     }
 }
 
@@ -48,8 +99,8 @@ impl ExactSizeIterator for CharRange {
         let start = self.low as u32;
         let end = self.high as u32;
         let naive_len = (end - start + 1) as usize;
-        if start <= BEFORE_SURROGATE && end <= AFTER_SURROGATE {
-            naive_len - (AFTER_SURROGATE - BEFORE_SURROGATE - 1) as usize
+        if start <= SURROGATE_RANGE.start && SURROGATE_RANGE.end <= end {
+            naive_len - SURROGATE_RANGE.len() as usize
         } else {
             naive_len
         }
--- a/unic/char/range/src/range.rs
+++ b/unic/char/range/src/range.rs
@@ -5,9 +5,13 @@ use step;
 const SURROGATE_RANGE: Range<u32> = 0xD800..0xE000;
 
 /// A range of unicode code points.
+///
+/// The members of this struct are public for const initialization by `chars!(..=)` only.
+/// They should be considered unstable private API that may change at any time.
+/// If you decide to use them anyway, make sure to note the safety notes.
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub struct CharRange {
-    /// The lowest uniterated character.
+    /// The lowest uniterated character (inclusive).
     ///
     /// Iteration is finished if this is higher than `high`.
     ///
@@ -16,12 +20,14 @@ pub struct CharRange {
     /// This is not guaranteed to always be a valid character. Check before using!
     /// Note that `high` _is_ guaranteed to be a valid character,
     /// so this will always be a valid character when iteration is not yet finished.
-    low: char,
+    #[doc(hidden)]
+    pub low: char,
 
-    /// The highest uniterated character.
+    /// The highest uniterated character (inclusive).
     ///
     /// Iteration is finished if this is lower than `low`.
-    high: char,
+    #[doc(hidden)]
+    pub high: char,
 }
 
 /// Constructors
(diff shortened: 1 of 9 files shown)


CAD97:
@behnam The actual implementation is ready for review. It's just the supporting chaff that I need to copy over -- the (half-)open constructors and the macro, along with test suite.

CAD97:
Table generation should probably use the `chars!( '\u{}' ..= '\u{}' )` syntax. It is a constant time expression, unlike the normal constructors, so can be used in the tables.

Reverse iteration is currently between 3 and 4 times slower than forward iteration. It must be slightly too complicated to inline in LLVM's eyes, since I did mark it `#[inline]` but it doesn't seem to have benefited in the same manner as forward iteration did. (In any case, forward iteration is the normal situation anyway.)

CAD97:
bors: try

unic/char/range/src/range.rs
```diff
@@ -0,0 +1,203 @@
...
+    /// Construct a fully open range of characters.
+    pub fn open(start: char, stop: char) -> CharRange {
+        let mut range = CharRange::closed(start, stop);
+        range.step_forward();
+        range.step_backward();
+        range
+    }
+
+    /// Construct a range of characters from bounds.
+    pub fn bound(mut start: Bound<char>, mut stop: Bound<char>) -> CharRange {
```
behnam: This is a very good way to keep a generic API here. ❤️ 

But, why do you expect the input to be `mut`? I think what you need is to take a readonly bound, and construct a value inside:
```
let start = if start == Bound::Unbounded { start = Bound::Included('\0') } else { start };
```

I guess this is the kind of issues we usually catch with unit tests. Want to add one for this?

CAD97: Of note: you can pass an imutable owned value here, because it's taken by value.

But this formulation makes more sense than what I have. 👍

CAD97:
[`std::ops::Range`](https://doc.rust-lang.org/std/ops/struct.Range.html) is an iterator. That is why I decided to in the "simple" version to make `CharRange` an iterator, and because it closer parallels the stdlib `Range`.

But on your argument towards separating `Range`/`RangeIter` doesn't work that well. Consider this minimal example ([playground link](https://play.rust-lang.org/?gist=8c148888b971de7af2bd3a74f14428a0&version=stable)):

```rust
#[derive(Copy, Clone, Debug, Eq, PartialEq)]
struct R {
    low: usize,
    high: usize,
}

impl R {
    fn contains(&self, item: usize) -> bool { self.low <= item && item <= self.high }
}

impl Iterator for R {
    type Item = usize;
    fn next(&mut self) -> Option<usize> {
        if self.low > self.high { None } else {
            let ret = self.low;
            self.low += 1;
            Some(ret)
        }
    }
}

fn main() {
    let range = R { low: 0, high: 20 };
    
    println!("{}", range.contains(10));
    for int in range { println!("{}", int); }
    println!("{}", range.contains(10));
}
```

This works. This works because [`impl<I: Iterator> IntoIterator for I`](https://doc.rust-lang.org/src/core/iter/traits.rs.html#249-257) does a move (or copy). You can move a normal non-mut range into a for loop and it will work. You cannot use it aftwerwards, though, because it is not `Copy`. Our `CharRange` is `Copy`, however, so moving it into the for loop makes a copy.

You can make an argument that the actual range object and the actual iteration are separate concerns, and should be handled by different structures. If I were designing a stdlib from scratch, I think I'd go the way of having a separate `Range` and `RangeIter` structures. However, for Rust, the ship has sailed with the design of `std::ops::Range`, and I think it's probably better to mimic the stdlib here.

CAD97:
I did a bit of research on why `Range` is not `Copy` when its index is.

rust-lang/rust#18045 rust-lang/rust#27186 rust-lang/rust#21846
https://botbot.me/mozilla/rust-libs/2015-05-19/?msg=39609682&page=1

It looks like `Range` is not `Copy` because it being having a copy iterator can be confusing:

https://github.com/rust-lang/rust/pull/27186#issuecomment-123390413

> These don't have it because they're iterators.
> The choice of removing Copy impls instead of adjusted for loop desugaring or linting was made to prevent this problematic case:
> 
> ```rust
> let mut iter = 0..n;
> for i in iter { if i > 2 { break; } }
> iter.collect()
> ```
> 
> Here `iter` is actually not mutated, but copied [if `Range` is `Copy`]. `for i in &mut iter` is required to mutate the iterator.

https://github.com/rust-lang/rust/issues/18045#issue-45808348

> ```rust
> let stream = "Hello, world!".chars().cycle();
> for _ in range(0u, 10) {
>     let chunk: String = stream.take(3).collect();
>     println!("{}", chunk);
> }
> ```

This is a footgun, and the reason the stdlib chose to make `Range` not `Copy`, because `Range` is an `Iterator`.

Note, having the range be directly iterable enables `(0..n).adapter().chain()`. But `(0..n).iter().adapter().chain()` is no worse in my opinion, and allows the `Range` object to be copy, and because we have `IntoIterator`, still be used in a for loop.

For our implementation, this also has the benefit of allowing us to fully expose the internals of the `Range` type without worrying about safety. `CharRange` can be a fully safe type with only `CharIter` having to (understandably) call unsafe code. (The unsafe iteration of `low` past `char::MAX` is an optimization, which you can see has a drastic effect when compared to reverse iteration. It's one less branch that has to be done on every step.)

This research into why `Range` is not `Copy` has moved me firmly onto the side of splitting up the range and iterator concepts properly.

CAD97:
This somehow sped up reverse iteration and slowed down forward. I have an idea of how reverse was sped up, and some things to try to get forward back below 1ms.

CAD97:
(See 7c0c5dc for a bit more on this last speedup.)

With that we're now consistently down below 1ms to iterate forward or backward accross the entire unicode range (on my machine) (with margin of error reaching slightly above). This is actually slightly better than the `char-iter` crate which is consistently just above 1ms (though within the reported margin of error of being below).

Forward and reverse iteration are within margin of error of each other in performance. I think this is a good indicator that neither path is favored over the other.

(Small note; if you just want to iterate over all characters it seems `(0..(char::MAX as u32+1)).rev().filter_map(char::from_u32)` is the fastest by a decent amount more than reported margin of error. 🤷‍♂️)

I'm happy with this. I'm ready to merge when you are @behnam.

unic/char/range/src/range.rs
```diff
@@ -1,39 +1,34 @@
...
 /// A range of unicode code points.
 ///
-/// The members of this struct are public for const initialization by `chars!(..=)` only.
-/// They should be considered unstable private API that may change at any time.
-/// If you decide to use them anyway, make sure to note the safety notes.
+/// The most idiomatic way to construct this range is through the use of the `chars!` macro:
+///
+/// ```
+/// # #[macro_use] extern crate unic_char_range;
+/// # use unic_char_range::*;
```
behnam: It's better to not hide the `extern crate` and `use` lines from the doc code blocks. (One empty line between them, plus an empty line before the actual code.)

unic/char/range/src/range.rs
```diff
@@ -1,39 +1,34 @@
...
+/// # fn main() {
+/// assert_eq!(chars!('a'..='z'), CharRange::closed('a', 'z'));
+/// assert_eq!(chars!('a'..'z'), CharRange::open_right('a', 'z'));
+/// assert_eq!(chars!(..), CharRange::all());
```
behnam: :heart:

unic/char/range/src/range.rs
```diff
@@ -0,0 +1,172 @@
...
+/// # }
+/// ```
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
```
behnam: I think derived `PartialEq` doesn't work for us here, because any case with `high < low` is considered empty for us, but `eq()` would return false depending on the numbers.

For rustc `RangeInclusive`, there's been a talk about having an `enum` for the type with an `Empty` variant. I don't think we want that here, but need to well-define emptiness.

Also a reminder that `CharRange` also needs an `is_empty()`, to not enforce `.len() == 0` to call-sites.

Another matter with empty case is the open question of if either of `low` and `high` matter in equality. IMHO, it doesn't matter, and we can have this:
```
fn eq(&self, other: Self) -> bool {
  if self.is_empty() {
    other.is_empty()
  } else {
    self.low == other.low && self.high == other.high
  }
}
```

What do you think?

unic/char/range/src/range.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    /// The highest character in this range (inclusive).
+    pub high: char,
+}
+
+/// Constructors
+impl CharRange {
+    /// Construct a closed range of characters.
+    ///
+    /// If `stop` is ordered before `start`, the resulting range will be empty.
```
behnam: I think this line can be moved to the type doc, with addition of `eq()` behavior. (And we can drop it from constructors.) (From d11n perspective, since the fields are `pub`, there shouldn't be a need to read the constructor docs to understand  the interpretation of the fields.)

unic/char/range/src/range.rs
CAD97: :+1: I'll include the lines.

unic/char/range/src/range.rs
CAD97: I actually did a bit of reading on `RangeInclusive` while making this and I _think_ the current design that's been landed on is having just two fields--a start and a stop--and creating `Step` fn to make a known-ended state.

And yes, it makes sense that two empty ranges no matter their internal state should compare the same.

unic/char/range/src/range.rs
CAD97: For some reason :heart: doesn't render on my machine so this looks like an empty comment 🙄

Since we have `CharRange::bound` I just realized we could also add `chars!(..'z')` and `chars!('a'..)`. I don't think those would be of much use though.

unic/char/range/src/range.rs
CAD97: :+1:

behnam:
bors: r+

PR #113, opened 2017-08-13
[char/property] Add CharProperty trait

behnam:
Add `CharProperty` trait as the core main contract for Character
Property definition, which in return requires defining the domain in
which the property is defined.

We should expand `char_property!` macro to accept these values and
implement the trait internally.

Also, we can see if we want to also enforce choosing a range for the
property. At this moment, `Age` yet doesn't have a contract trait.

### DIFF
--- a/unic/char/property/src/domain.rs
+++ b/unic/char/property/src/domain.rs
@@ -16,6 +16,19 @@ use std::fmt::{Debug, Display};
 use std::hash::Hash;
 
 
+/// A Character Property, defined for some or all Unicode characters.
+pub trait CharProperty: PartialCharProperty {
+    /// Get property abbreviated name
+    fn prop_abbr_name() -> &'static str;
+
+    /// Get property long name
+    fn prop_long_name() -> &'static str;
+
+    /// Get property human-readable name
+    fn prop_human_name() -> &'static str;
+}
+
+
 /// A Character Property defined for some characters.
 ///
 /// Examples: *Decomposition_Type*, *Numeric_Type*
@@ -30,6 +43,8 @@ pub trait PartialCharProperty: Copy + Debug + Display + Eq + Hash {
 /// Examples: *Age*, *Name*, *General_Category*, *Bidi_Class*
 // Because of rustc bug, we cannot rely on inheritance for the moment.
 // See: <https://github.com/rust-lang/rust/issues/43777>
+// See: <https://github.com/rust-lang/rust/issues/43784>
+// TODO: Drop extra super-traits after the bugs are fixed.
 //pub trait CompleteCharProperty: PartialCharProperty + Default {
 pub trait CompleteCharProperty
     : PartialCharProperty + Copy + Debug + Display + Eq + Hash + Default {
--- a/unic/char/property/Cargo.toml
+++ b/unic/char/property/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["The UNIC Project Developers"]
 repository = "https://github.com/behnam/rust-unic/"
 license = "MIT/Apache-2.0"
 keywords = ["text", "unicode", "character-property"]
-description = "UNIC - Unicode Characters - Character Property"
+description = "UNIC - Unicode Characters - Character Property taxonomy, contracts and build macros"
 categories = ["text-processing", "parsing"]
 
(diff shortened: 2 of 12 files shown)


behnam:
Values are filled in based on <http://www.unicode.org/Public/UCD/latest/ucd/PropertyAliases.txt>

unic/char/property/src/domain.rs
```diff
@@ -16,6 +16,19 @@ use std::fmt::{Debug, Display};
 use std::hash::Hash;
 
 
+/// A Character Property, defined for some or all Unicode characters.
+pub trait CharProperty: PartialCharProperty {
```
CAD97: Every property in [the Property Index](http://unicode.org/reports/tr44/#Property_Index_Table) has an entry in [PropertyAliases.txt](http://www.unicode.org/Public/UCD/latest/ucd/PropertyAliases.txt).

I think we should make `CharProperty` the supertrait and `PartialCharProperty`, `CompleteCharProperty`, `EnumeratedCharProperty`, and `NumericCharProperty` just extend from that.

We can also drop `Partial` as a parent of `Complete` if we do so, which was only done such that there _would_ be a common parent to have as a supertrait of `NumericProperty`/`EnumeratedProperty`.

(Note: I haven't considered Unihan, which has more than the 15 properties that have an entry in PropertyAliases.txt. However, a cursory examination shows that all normative properties have an alias in said document. And we could always just return the same string from both if there's only one.)

CAD97: So:

```rust
trait CharProperty: Copy + Debug + Display + Eq + Hash;
trait PartialCharProperty: CharProperty;
trait CompleteCharProperty: CharProperty + Default;
trait EnumeratedCharProperty: CharProperty + Sized;
trait NumericCharProperty<Value>: CharProperty;
```

CAD97: We could even keep the `impl PartialCharProperty for CompleteCharProperty` if we wanted, but the advantage of having them be disjoint is that importing both into the namespace will still allow unambiguous `::of()` calls.

behnam:
About the macro design, I was thinking something more similar to how we assign these strings to variants. Like this:

```
char_property! {
    abbr => "myprop";
    long => "MyProp";
    human => "MyProp";

    pub enum MyProp {
        /// Required
        AbbrVariant {
            abbr => AV,
        }
        /// Required
        LongVariant {
            abbr => LV,
            long => Long_Variant,
        }
        /// Required
        DisplayVariant {
            abbr => DV,
            display => "The one and only DISPLAY VARIANT!",
        }
        /// Required
        EmptyVariant {
            abbr => EV,
        }
    }

    pub mod abbr_names for abbr;
    pub mod long_names for long;
}
```

About if abbr/long is appropriate for all properties, I agree that we can just use whatever's available if something specific is not defined in the source. I believe UAX#44 promises to define all of these, though.

behnam:
About the taxonomy, here's the constraint I'm trying to hold: *Whenever a Character Property is defined, there's a clear domain for the mapping, as well as a clear range.*

As noted above, I haven't got to the range part yet. But that's going to be similar to the domain part, below.

So, how I'm enforce domain definition is to have `trait CharProperty : PartialCharProperty`. If we make it the other way here, then any user can `impl CharProperty` without any need to have Partial/Complete domain, and basically not satisfy the invariants.

For the domain, it actually makes more sense to not have an abstract root trait, because Complete is a subset of Partial. Having `trait Complete : Partial` basically allows writing any code once when it can work on both. (And write it once when it can only work on Complete, obviously.) So, I think we actually ended up with a very good design here.

About the range, it may make more sense to have an abstract root, which would be needed by `CharProperty`. I'm not sure yet, and need to work on it more.

behnam:
> We could even keep the `impl PartialCharProperty for CompleteCharProperty` if we wanted, but the advantage of having them be disjoint is that importing both into the namespace will still allow unambiguous `::of()` calls.

Actually, since we implement `of()` on the main type, that takes priority over traits and there won't be any conflicts. But, unfortunately, that's one of things we cannot enforce on other users via type tools. So, I guess we need to document this expectation in the docs.

CAD97:
bors: r+

PR #115, opened 2017-08-13
[char/property] Re-org code and update docs

behnam:


unic/char/property/src/lib.rs
```diff
@@ -12,14 +12,19 @@
 
 #![forbid(unsafe_code, unconditional_recursion, missing_docs)]
 
-//! # UNIC — UCD — Character Age
```
CAD97: Wait what? o_0 Why was it that before? Who missed that in review? 😆

behnam: Because I copy-pasted and we never looked at it again! 😃

PR #117, opened 2017-08-13
[char/property] Add long_name()/human_name() to EnumeratedCharProperty

behnam:
Add `long_name(&self)`/`human_name(&self)` to the
`EnumeratedCharProperty` contract and add values to the implementing
types.

Also needs to be implemented by `char_property!` macro.

These name methods *may* not belong to *Enumerated* contract, but this
is the only contract we have right now that provides them, so let's
start with having them here.

Bascially, looks like we have to major categories of property range
types: those with pre-defined set of possible values (*Enumerated*, and
possibly other ones), and those with numeric/complex values, which don't
have *named* variants for all of them, which would result in
name-generation to be runtime and not compile time.

For the second group, we can add another parent contract to implement
these methods as `fn *_name(&self) -> String`, which are clearly more
expensive and may not be used when `&'static str` is accepted.

This way, we keep the API similar enough to remember, but also not
enforcing runtime overhead.

### DIFF
--- a/unic/char/property/src/macros.rs
+++ b/unic/char/property/src/macros.rs
@@ -369,6 +369,14 @@ macro_rules! __char_property_internal {
                 }
             }
 
+            fn long_name(&self) -> &'static str {
+                "FIXME"
+            }
+
+            fn human_name(&self) -> &'static str {
+                "FIXME"
+            }
+
             fn all_values() -> &'static [$name] {
                 const VALUES: &[$name] = &[
                     $($name::$variant,)+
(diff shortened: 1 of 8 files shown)


behnam:
bors: r+

CAD97:
I would use [`Cow<'static, str>`](https://doc.rust-lang.org/std/borrow/enum.Cow.html) for `NumericProperty`. [CCC](http://unicode.org/reports/tr44/#Canonical_Combining_Class_Values) has several names which are known at static time and as such can be given as borrowed `&'static str` rather than generated. It should also probably be `Option` as some CCC values (e.g. ccc=255) don't have names: the procedural "CCCindex" name only applies to values ccc=10 through ccc=199.

PR #118, opened 2017-08-13
[ucd] Drop `use ...::*` lines outside of data or tests

behnam:
the `use ...::*` is considered a bad practice and should be avoided as
much as possible, or limited to the scope where it's needed.

A major case we have needed them in UCD components is when loading data
from `rsv` files, where we don't want to prefix values in every single
line with the type name. To limit the scope of the wildcard usage, I
have moved these to a local `mod data`.

### DIFF
--- a/unic/ucd/category/src/category.rs
+++ b/unic/ucd/category/src/category.rs
@@ -254,16 +254,18 @@ impl Default for GeneralCategory {
 }
 
 
-use self::abbr_names::*;
+mod data {
+    use super::abbr_names::*;
+    pub const GENERAL_CATEGORY_TABLE: &'static [(char, char, super::GeneralCategory)] =
+        include!("tables/general_category.rsv");
+}
 
-const GENERAL_CATEGORY_TABLE: &'static [(char, char, GeneralCategory)] =
-    include!("tables/general_category.rsv");
 
 
 impl GeneralCategory {
     /// Find the `GeneralCategory` of a single char.
     pub fn of(ch: char) -> GeneralCategory {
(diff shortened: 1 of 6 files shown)


unic/ucd/category/src/category.rs
```diff
@@ -321,41 +325,49 @@ impl GeneralCategory {
 impl GeneralCategory {
     /// `Lu` | `Ll` | `Lt`  (Short form: `LC`)
     pub fn is_cased_letter(&self) -> bool {
+        use self::abbr_names::*;
```
CAD97: I personally think that this specific use case is easier to read with the localized `use abbr_names::*`, because it then mirrors the source almost exactly.

Still, the title of this PR is "Drop `use ...::*` lines outside of data or tests", and this is neither data nor a test, so it either needs to drop the `use ::*` or I would like to see a reasoning somewhere in the text of the PR saying why these get to keep their `use ::*`.

CAD97:
As a side note, this touches a _lot_ of code which is also touched by my current work changing all of the properties to use the property macro. It would _probably_ be easier (and less error-prone) if I can land that first then we rebase this smaller change.

unic/ucd/category/src/category.rs
behnam: I thought I explained it with the "or limited to the scope where it's needed" phrase. This is basically limiting the use of `*` to a small block of code, instead of one whole `impl`.

CAD97: OK, yeah, that does cover this use.

behnam:
Rebased now. @CAD97, bors, if you're okay?

PR #119, opened 2017-08-14
[char/property] [ucd] Macro update and use

CAD97:
I can submit a separate PR after 31e6582 if we want to pull the edited macro and the application separately.

The diff is kind of opaque because this touches so much. You're probably better off just comparing before/after the PR without looking at the diff output.

This PR is submitted in a ready-to-merge (after review) state rather than my more usual track-final-polish.

This will require a rebase of #118 as it touches many of the same parts of the code.

### DIFF
--- a/unic/bidi/tests/conformance_tests.rs
+++ b/unic/bidi/tests/conformance_tests.rs
@@ -13,6 +13,7 @@
 
 
 extern crate unic_bidi;
+extern crate unic_char_property;
 
 
 use unic_bidi::{format_chars, level, BidiClass, BidiInfo, Level};
@@ -279,33 +280,9 @@ fn gen_char_from_bidi_class_abbr(bidi_class_abbr: &str) -> char {
 
 #[test]
 fn test_gen_char_from_bidi_class() {
-    use self::bidi_class::abbr_names::*;
+    use unic_char_property::EnumeratedCharProperty;
 
-    for &class in &[
-        AL,
-        AN,
-        B,
-        BN,
-        CS,
-        EN,
-        ES,
-        ET,
-        FSI,
-        L,
-        LRE,
-        LRI,
-        LRO,
-        NSM,
-        ON,
-        PDF,
-        PDI,
-        R,
-        RLE,
-        RLI,
-        RLO,
-        S,
-        WS,
-    ] {
+    for &class in BidiClass::all_values() {
         let bidi_class_abbr = class.abbr_name();
         let sample_char = gen_char_from_bidi_class_abbr(bidi_class_abbr);
         assert_eq!(BidiClass::of(sample_char), class);
--- a/unic/bidi/tests/conformance_tests.rs
+++ b/unic/bidi/tests/conformance_tests.rs
@@ -15,7 +15,7 @@
 extern crate unic_bidi;
 
 
-use unic_bidi::{bidi_class, format_chars, level, BidiClass, BidiInfo, Level};
+use unic_bidi::{format_chars, level, BidiClass, BidiInfo, Level};
 
 
 const BASIC_TEST_DATA: &'static str = include_str!("../../../data/ucd/test/BidiTest.txt");
--- a/unic/ucd/normal/src/decomposition_type.rs
+++ b/unic/ucd/normal/src/decomposition_type.rs
@@ -12,55 +12,155 @@
 
 //! Accessor for *Decomposition_Type* (dt) property
 
-use std::fmt;
-
 use unic_utils::CharDataTable;
-use unic_char_property::{CharProperty, EnumeratedCharProperty, PartialCharProperty};
+use unic_char_property::PartialCharProperty;
 
 use composition::{canonical_decomposition, COMPATIBILITY_DECOMPOSITION_MAPPING};
 use hangul;
 
+char_property! {
+    /// Represents the Unicode character
+    /// [*Decomposition_Type*](http://www.unicode.org/reports/tr44/#Decomposition_Type) property.
+    ///
+    /// * <http://www.unicode.org/reports/tr44/#Character_Decomposition_Mappings>
+    pub enum DecompositionType {
+        abbr => "dt";
+        long => "Decomposition_Type";
+        human => "Decomposition Type";
 
-/// Represents the Unicode character
-/// [*Decomposition_Type*](http://www.unicode.org/reports/tr44/#Decomposition_Type) property.
-///
-/// * <http://www.unicode.org/reports/tr44/#Character_Decomposition_Mappings>
-#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
-#[allow(missing_docs)]
-pub enum DecompositionType {
-    Canonical,
-    Compat,
-    Circle,
-    Final,
-    Font,
-    Fraction,
-    Initial,
-    Isolated,
-    Medial,
-    Narrow,
-    Nobreak,
-    None,
-    Small,
-    Square,
-    Sub,
-    Super,
-    Vertical,
-    Wide,
-}
+        #[allow(missing_docs)]
+        Canonical {
+            abbr => Can,
+            long => Canonical,
+            human => "Canonical",
+        }
 
+        /// Otherwise unspecified compatibility character
+        Compat {
+            abbr => Com,
+            long => Compat,
+            human => "Unspecified",
+        }
 
-impl CharProperty for DecompositionType {
-    fn prop_abbr_name() -> &'static str {
-        "dt"
+        /// Encircled form
+        Circle {
+            abbr => Enc,
+            long => Circle,
+            human => "Encircled",
+        }
+
+        /// Final presentation form (Arabic)
+        Final {
+            abbr => Fin,
+            long => Final,
+            human => "Arabic Final",
+        }
+
+        /// Font variant (for example, a blackletter form)
+        Font {
+            abbr => Font,
+            long => Font,
+            human => "Font Variant",
+        }
+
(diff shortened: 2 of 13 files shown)


CAD97:
bors: try

unic/bidi/tests/conformance_tests.rs
```diff
@@ -279,33 +280,9 @@ fn gen_char_from_bidi_class_abbr(bidi_class_abbr: &str) -> char {
...
-        R,
-        RLE,
-        RLI,
-        RLO,
-        S,
-        WS,
-    ] {
+    use unic_char_property::EnumeratedCharProperty;
+
+    for &class in BidiClass::all_values() {
```
behnam: Yaay! Finally!

unic/ucd/normal/src/decomposition_type.rs
```diff
@@ -12,228 +12,180 @@
...
+            abbr => Nb,
+            long => Nobreak,
+            human => "No-Break",
+        }
 
-    fn human_name(&self) -> &'static str {
-        self.human_name()
-    }
-}
+        // TODO: Is this used?
```
behnam: No, instead of supporting `None` variant in the type, we made the type Partial, using `Option::None` for this case. So, we can drop this variant here.

Basically, the idea is to not create a new `None` type and build tooling around it while `Option::None` does the job and provides the same semantics. I'm going to update the *Unicode API* doc to explain these, as soon as we're done with the work here.

unic/ucd/normal/src/decomposition_type.rs
```diff
@@ -12,228 +12,180 @@
...
-    pub use DecompositionType::Sub;
-    pub use DecompositionType::Super as Sup;
-    pub use DecompositionType::Vertical as Vert;
-    pub use DecompositionType::Wide;
-}
+        /// CJK squared font variant
+        Square {
+            abbr => Sqr,
+            long => Square,
+            human => "Squared",
```
behnam: I believe we can expand all the `human` lines here to be more descriptive. For example, "Squared" here is not actually that useful. The point is that it's a "CJK squared font variant".

I think it's safe to go with the full text of the *Description* column in the table. Maybe only drop the *form* words and the parenthesis, which are more of a side note.

We can update this in a separate PR, but noticed some changes from the basic words and this item specifically caught my eye.

unic/ucd/normal/src/decomposition_type.rs
```diff
@@ -12,228 +12,180 @@
...
+            abbr => Nar,
+            long => Narrow,
+            human => "Narrow",
+        }
 
-    fn long_name(&self) -> &'static str {
-        self.long_name()
-    }
+        /// No-break version of a space or hyphen
+        NoBreak {
```
behnam: Our general approach is to take the long name and drop the underscores. I think special-casing `NoBreak` does have much gain to worth it. What do you think?

unic/ucd/normal/tests/conformance_tests.rs
```diff
@@ -67,7 +67,7 @@ fn test_decomposition_type_conformance() {
                     .unwrap_or(low);
                 low..(high + 1)
             };
-            let exp_dt = Some(get_dt_from_name(fields[1]));
+            let exp_dt = Some(fields[1].parse().expect("Invalid Decomposition Type"));
```
behnam: Yaaay!

unic/ucd/normal/src/decomposition_type.rs
CAD97: 👍

I had the `None` type here because it was in the version before migrating to the macro. But now it's gone.

unic/ucd/normal/tests/conformance_tests.rs
CAD97: There are probably other places where the new APIs can clean up tests; I didn't go over everything with a fine tooth comb. I only really got the ones in this PR because they popped up as a compilation error.

unic/ucd/normal/src/decomposition_type.rs
CAD97: I've changed this one to `"CJK Squared"` as it was the most ambiguous.

unic/ucd/normal/src/decomposition_type.rs
CAD97: (I think you missed a few words, I can't tell if that's slightly towards or against.)

Here's my main logic:

- The [formatting tag](http://unicode.org/reports/tr44/#Formatting_Tags_Table) is `<noBreak>`
  - (This would allow dropping the `to_lowercase` call in `unic-gen::generate::ucd::normal::CompatibilityDecompositionData::emit`, though I haven't here and used the `long_names` module instead)
- It's described as "No-break" (two words)
- My spell checker gets annoyed about "Nobreak"
  - It fits Rust conventions of `CamelCase`; `Nobreak` feels like `Runoncase` because conceptually it is the "no-break" type and not the "nobreak" type

I thought I'd change this here and see what your opinion was here (good on you to catch it; I was actually going to mention it but I forgot), as now we have access to it as `Nobreak` in `mod long_names`.

unic/ucd/normal/src/decomposition_type.rs
behnam: On, I must have missed it when copy-pasting the list!

unic/ucd/normal/src/decomposition_type.rs
behnam: Oops! I meant "does *not* provide much gain".

Yeah, totally agree with all the arguments here. But, also, similar to some other cases before, we are going to need a clear way to explain it as a generic overall rule.

For this, the fact that the spec and data files use `noBreak` is not enough to define the API based on. I mean, the more formal form, `PropertyAliases.txt`, is more important, and I want us to be able to describe the decisions based on that more formal data.

But agreed, this one term is in such a messy situation now. Unfortunately there's a tendency to not fix these things in the updates.

To be clear, I wish we can come up with a policy to justify it. I just can't think of any yet.

CAD97: I'm pretty sure the UCD will never change the representation in a file to support old parsers for as long as possible. By the spec, comparison of property keys and values should be case insensitive and ignore `_` and ` `. (So "`NOBREAK`", "`NO_BREAK`", "`n  o _    _ bre  ____AK`" are all equivalent values.)

In #58 we discussed a similar case with naming `ccc=204` and the other classes with a description with no name. There we reached a rough consensus at

> @cad97
> The guideline I used here was basically  
> "The Unicode document provides a description for the (unnamed) value, which follows a pattern of descriptions for other (named) values, from which the description => name transformation can be trivially derived"
>
> @behnam 
> Sound reasonable, specially since it's not the [name in the UCD name alias module], but a local naming based on the [UCD name].

I think we can expand that to slightly modifying slightly-out-of-line UCD naming by a similar rule:

> [PROPOSED]
> Naming of enum variants (for `EnumeratedCharProperty`s) and variant-like constants (for other properties) follows Rust naming conventions for enum variants (namely, `CamelCase`). Namely, Unicode "Long Aliases" of the format `Long_Name` (sourced from PropertyValueAliases) will have their underscores removed to create a `CamelCase` identifier. If other appearances of the data in normative sources use a form not listed in PropertyValueAliases.txt, the "word breaking" (in whatever form: `snake_case`, `lowerCamel`, `CamelCase`, `SCREAMING_CASE` `Sentence structure`) of that format will be considered; and if it is deemed to have more natural breaks, those will be used instead.

CAD97: Second proposal, simple, permissive, but guaranteed conformant:

> Rust enum variant names are picked based on Rust naming conventions. A default identifier matching the formal long alias under [UAX44-LM3](http://unicode.org/reports/tr44/#Matching_Symbolic) will be chosen and remain consistent under Rust API compatibility rules. Aliases matching the formal long and abbreviated aliases will be provided when possible by sister modules to the types in question.

behnam:
Okay, there are two different issues here:

1) Matching strings, for which we have `FromStr()` implemented, which now looks like it better has UAX44-LM3 fully implemented.

2) API design for Rust, and a way to make it learnable. Having the Rust names here UAX44-LM3-compatible with long-names is a very good thing, and we should be able to guarantee it.

   As good as the rules above are, there's still the fact that writing `Nobreak` in Rust won't match what we name as `NoBreak`.

Now, maybe we can just rely on the nice rustc errors on mismatches, and say that we will use a name UAX44-LM3-compatible with the long-names, with some changes here and there to improve readability and usage.

What do you think?

behnam:
Btw, let's land the PR without None, and we can update no-break after discussion.

CAD97:
(A [regex search](https://regex101.com/r/F4EflY/1) shows only `Nobreak` which really would want to be modified. Yay for outlier data points?)

Technically `FromStr` isn't quite UAX44-LM3 compliant -- it only ignores case and not `[_- ]`. (It would not be too difficult to do so however, so long as we make sure to have a test for the exception to ignoring `-`.)

So revert the variant to `Nobreak` for now? @behnam

behnam:
Yeah, what I meant that I think we can/should expand `FromStr`. That way, we can keep the API more friendly and also relax the rewriting in unic-gen.

About no-break, I think `NoBreak` should be fine. Let's go ahead with that, and we try to make it fit. We just better decide on it before the next release. (5 more weeks.)

CAD97:
Ready to merge when you are, captain! 😉

behnam:
Uh, I didn't notice you dropped `None` already. Let's go then!

behnam:
bors: r+

Issue #120, opened 2017-08-14
UCD Age: Custom enum type or Option<UnicodeVersion>?


behnam:
Right now, we have Age as this:

<https://github.com/behnam/rust-unic/blob/72ca9a893373f1857bc6ab6440389dfdceea6f13/unic/ucd/age/src/age.rs#L36-L42>

1. It's nice to have this `CompleteCharProperty`, as it gives meaningful API like `is_assigned()` and `is_unassigned()`.

2. Another option is to convert this to a `PartialCharProperty`, which returns `None` if the char is `Unassigned`, and `Some(UnicodeVersion)` otherwise.

3. One more option would be to have it as `PartialCharProperty`, but the return type will be more following the UAX#44 spec: only having major and minor numbers. Basically, return value would be `Some(Age)`, with `Age` being similar to `UnicodeVersion`, but without minor number. Then, we can have transformations between `Age` and `UnicodeVersion`.

Option (3) makes it much more like other types in UCD, option (2) is a bit farther, and option (1), which is the current one, is the farthest from general approach here.

BUT, there's also the fact that option (1) looks nicer and more organic comparing to the other ones.

I'm filing this mostly because I need to figure out a way to provide a property-range contract for `Age`, and would be great if it's not just an empty contract.

Any ideas?

CAD97:
This is not an easy question.

In the best case scenario, only _valid_ ages (currently [1.1, 2.0, 2.1, 3.0, 3.1, 3.2, 4.0, 4.1, 5.0, 5.1, 5.2, 6.0, 6.1, 6.2, 6.3, 7.0, 8.0, 9.0, 10.0]) are valid `Age`s. [Age is a Category type after all.](http://unicode.org/reports/tr44/#Age) (That would suggest it should be a `EnumeratedCharProperty`. It even has defined aliases in ProperyValueAliases.txt!)

So you'd have a `impl From<Age> for UnicodeVersion` and a `impl TryFrom<UnicodeVersion> for Age`.

Unfortunately, in order to create `Age`s for our components at constant time, (at least until `const fn` and `static_assert!` stabilize) the field(s) of `Age` have to be public. This means we cannot guarantee that our `Age` is valid accross the above domain.

But! If we treat it like the category type it is, but use a more rich typing behind the scenes, we might get something like the following:

```rust
#[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash)]
pub struct Age {
    version_major: u16,
[...]

PR #123, opened 2017-08-14
[ucd/age] Change Age type to struct

behnam:
The *Age* property was implemented as an `enum` type which was closely
mirroring the core `Option` type. Now that we have a clear definition of
`Partial` properties, it makes sense to update `Age` to be a Custom
property type (`struct`) with *actual* `UnicodeVersion` value.

But, there was a need to update one thing about the Age, it's ordering!
Previously, Age was ordered in the same direction as `UnicodeVersion`,
and `Unassigned` value was larger than any `Assigned(UnicodeVersion)`.
That didn't make much sense in common understanding of *age*, but was
consistent with `UnicodeVersion`.

Now, because of the fact that we cannot change the ordering of `Option`
variants (`None` and `Some(UnicodeVersion)`, in this case), and also to
put the meaning of comparison closer to the common understanding of the
word, we reverse the direciton of ordering for `Age`. Therefore, `Age`
value for character `'A'` is *greater than or equal to* any other
character in Unicode, assigned or unassigned.

Closes <https://github.com/behnam/rust-unic/issues/120>


CAD97:
I would like to say that having `Unassigned > Assigned` is ideal.

<http://unicode.org/reports/tr44/#Character_Age>

> The typical use case for the Age property in regular expressions is to search for all characters that were present in a given version. For this reason, an expression such as "\p{age=V3_0}" is exceptionally defined to match all of the code points assigned in Version 3.0—that is, all the code points with a value less than or equal to the value 3.0 for the Age property, rather than just the subset of those code points with the value 3.0. This interprets "\p{age=V3_0}" as the set of all characters assigned as of Unicode 3.0, rather than as just the set of characters added to Unicode 3.0 subsequent to the prior version. For more information, see Unicode Technical Standard #18, "Unicode Regular Expressions" [[UTS18]].

So the filter for this behavior is `Age::of(char) <= Age(UnicodeVersion::parse("3.0.0")` if Unassigned > Assigned.

The key point here is to realize that the `Age` property is not some how long the character has been around, but the version in which the codepoint was assigned.

[UTS18]: http://unicode.org/reports/tr41/tr41-21.html#UTS18

CAD97:
That said, (and see #120 for much more said), I feel using `Option` rather than re-implementing the concept is best. Anybody programming in Rust already should understand the ordering of `Option::None` versus `Option::Some` as well, so this avoids surprises there by using the stdlib type.

unic/ucd/normal/src/canonical_combining_class.rs
```diff
@@ -53,7 +53,7 @@ impl TotalCharProperty for CanonicalCombiningClass {
 impl NumericCharProperty<u8> for CanonicalCombiningClass {
     /// Get numeric value for character property value
     fn number(&self) -> u8 {
-        self.number()
+        Self::number(self)
```
CAD97: Wait. How does this work? Is this not self-recursive and should it not just be `self.0`?

gen/src/generate/ucd/age.rs
```diff
@@ -74,7 +74,7 @@ impl FromStr for AgeData {
                     age_data.insert(
                         char,
                         format!(
-                            "Assigned(UnicodeVersion {{ major: {}, minor: {}, micro: {} }})",
+                            "UnicodeVersion {{ major: {}, minor: {}, micro: {} }}",
```
CAD97: This feels so much nicer. 🎉

behnam: Yeah! Actually, I think we should do the same for all non-named properties. Doing so, we don't have to make the struct fields public at all!

unic/ucd/age/src/age.rs
```diff
@@ -10,36 +10,37 @@
...
-/// The *earliest* value for this property is `UnicodeVersion { major: 1, minor: 1, micro: 0 }`,
-/// because of the massive changes for the merger of the Unicode Stanrda with ISO 10646.
+/// Unicode versions with character *assignement* always have the Micro (Update) version value
+/// of zero (`0`). Therefore, all `UnicodeVersion` values return from `Age` will have their `macro`
+/// field as `0`.
 ///
-/// The *latest* value for this property is always equal to or less than `UNICODE_VERSION`. (Only
-/// not equal when `UNICODE_VERSION` has non-zero *micro* value.)
+/// The *earliest* (largest) value for this property is `UnicodeVersion { major: 1, minor: 1, micro:
+/// 0 }`, because of the massive changes for the merger of the Unicode Stanrda with ISO 10646.
```
behnam: Thanks! Fixed here and in other copy.

unic/ucd/normal/src/canonical_combining_class.rs
behnam: If a type has a method, that always takes precedence. So, calling into such a method from a trait is safe, as long as you remember to implement the function. With the right lint warn enabled, we can easily catch the problem in compile time, if happens. (Which actually did while I was expanding the contracts! :D)

behnam:
bors: r+

PR #124, opened 2017-08-14
[char/property] Rename to TotalCharProperty

behnam:
Rename the trait to match the mathematical term for these kinds of
relationships.


behnam:
bors: r+

PR #125, opened 2017-08-15
[ucd/name] Add minimal name component

CAD97:
Replaces #103 

This is meant to be minimal. We can and should expand this further.

To get it written down, a good expansion is a `impl PartialEq<&str>` following UAX44-LM2.


unic/ucd/name/src/name.rs
```diff
@@ -0,0 +1,41 @@
...
+
+
+#[derive(Copy, Clone, Debug, Ord, PartialOrd, Eq, PartialEq, Hash)]
+pub struct Name {
+    pieces: &'static [&'static str],
+}
+
+impl Name {
+    pub fn of(ch: char) -> Option<Name> {
+        data::NAMES.find(ch).map(|&pieces| Name { pieces })
```
behnam: Wondering, since we have this `map()` call in `Age` and now here, does it come with a runtime cost? I'm guessing it should be reduced well in `release` builds that it doesn't matter, but I'm not sure if we need to actually test it before using it for the char-props that are more critical, like normal and bidi components. What do you think?

unic/ucd/name/src/name.rs
```diff
@@ -0,0 +1,41 @@
...
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+
+use std::fmt;
+
+use unic_utils::CharDataTable;
+
+
+#[derive(Copy, Clone, Debug, Ord, PartialOrd, Eq, PartialEq, Hash)]
```
behnam: What does `PartialOrd` do here? Does it really implement an ordering that derefs all the strings and comparing them using `str`'s ordering?

I'm getting more conscious about deriving `*Ord` traits now. They are not in the contracts, and maybe we should actually not derive them unless we know what we want and we add unit tests for them?

Also, this now reminds me that this module needs at least a minimal unit test to check a couple of basic entries and default values (`None`).

I'll work on the contract part after landing, so don't worry about that part.

unic/ucd/name/src/name.rs
CAD97: `Option::map` is what we would do anyway:

<https://github.com/rust-lang/rust/blob/d21ec9b4efd1da012979b050bc0a0426fe45fcdf/src/libcore/option.rs#L378-L401>

```rust
    pub fn map<U, F: FnOnce(T) -> U>(self, f: F) -> Option<U> {
        match self {
            Some(x) => Some(f(x)),
            None => None,
        }
    }
```

I agree we'd want to test it, but for the structures that are just simple wrappers, it should be a noop anyway.

unic/ucd/name/src/name.rs
CAD97: It's a lexographic ordering of the slice: <https://doc.rust-lang.org/src/core/slice/mod.rs.html#2499-2503>

I'll add a couple basic tests in.

CAD97: In the case of the name property, it's defined to return a string. Here I've returned a custom structure such that we don't have to allocate extra space for the `String`, but the intent is it should act like a `&str` where possible.

CAD97:
Is there a simple way to see how much space a table adds to the crate? `mem::size_of` won't work because most of the data is in a slice somewhere in the static data.

behnam:
> Is there a simple way to see how much space a table adds to the crate? `mem::size_of` won't work because most of the data is in a slice somewhere in the static data.
> 
> The tables for the character names are going to be large, there is no way around that. We should probably add a note that the crate `unic-ucd` exists for convenience, but should not be used, preferring to instead use each specific subcrate which is required.

I've been wondering about this for a while, as well. I think we need to monitor the size of both *crate file* (the archive file with `.crate` suffix, which is what gets uploaded to and downloaded from the registry) and the size of the dynamic library, which depends on the platform.

I don't know how else people measure these outside of a larger platform. I wish there was a 3rd-party cargo command (with good understanding of workspaces) to provide a summary to compare easily.

behnam:
bors: r+

behnam:
Right now on my Mac machine, it adds about 10-20 seconds of build time and the dynamic library file (`rlib`) is 26 MB.

PR #126, opened 2017-08-15
GitHub: Ignore generated files

CAD97:
GitHub thinks the repository is 53.7% RenderScript and 45.5% Rust. Oops!

RenderScript

This just tells GitHub/linguist to ignore any files matching `**/tables/*` as generated code.

Another option is the following .gitattributes:

```gitattributes
*.rs linguist-language=Rust
*.rsv linguist-generated=true
```

<https://github.com/github/linguist#troubleshooting>

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+**/tables/* linguist-generated=true


behnam:
Huh! The RenderScript thing is new. It was working fine until very recently.

Anyways, yeah, this PR is in the right direction. We can expand if needed.

Also, it just reminded me about auto-generated `.rs` files. I intentionally changed the suffix to prevent formatting on them. I think we can pick a similar suffix for the non-value include files, too. Unless `rustfmt` has a good way to be suppressed for the same pattern.

behnam:
bors: r+

CAD97:
I suspect the reason that it got picked up as RenderScript is https://github.com/github/linguist/issues/3600

Basically the `.rs` generated file in ucd/name isn't rusty enough and linguist picked RenderScript.

If `rsv` == "Rust Value", maybe it could be `rsd` ("Rust Declarations")? In any case, it definitely should not be `rsv` as it is not a value, but rather a list of constants.

CAD97:
By https://github.com/rust-lang/rfcs/pull/2108 it'll have to change or move out of the `src` directory anyway.

behnam:
Hum... it's a valid point that we can now move auto-generated files out of `src/`. I think we shouldn't put them under components' `data/`, because I consider `data` to have source data, not auto-generated. Should we just keep the `tables/`, as in almost all cases these are data of the form of tables (except versions)?

CAD97:
An option is `generated-src` next to `src`. (or `generated`.) Or `tables` next to `src`. I think just having the folder be called `tables` is fine.

behnam:
Okay, the simplest solution wins: let's move them out of `src` without rename or anything. Just updating all the paths with dropping `src/`.

I won't be able to do it this evening, but if you like to take care of it, I can stamp.

PR #132, opened 2017-08-31
Fix compile on nightly

CAD97:
See https://github.com/rust-lang/rust/issues/44208 for context


CAD97:
bors: r+

behnam:
Thanks for the fix, @CAD97! I noticed this today, too, and was wondering how it wasn't the case before. The `CharRange` fix will also help, but good to have a hotfix right now. :)

PR #136, opened 2017-09-04
Turn CharDataTable into a dedicated type using CharRange

CAD97:
r? @behnam 

Since everything needs access to `chars!` now there are some wide-reaching changes here. All the tests pass though, and I think I got everything.

The traits/fn in `unic_gen::generate::char_property` feel wrongly named now, any ideas on better naming?


CAD97:
@behnam I don't like to be that guy pinging you multiple times but... r? I don't want to merge it myself without an OK from you.
This is a prerequisite for adding in more components.

unic/char/range/src/range.rs
```diff
@@ -148,6 +149,24 @@ impl CharRange {
         self.low <= ch && ch <= self.high
     }
 
+    /// Determine the ordering of this range and a character.
+    ///
+    /// # Panics
+    ///
+    /// Panics at runtime _in debug builds only_ if the range is empty.
+    /// In release builds, the ordering will never be `Equal`,
```
behnam: IIUC, this claim is based on how this function is used, by not how it's implemented. I think it needs to be an assertion in some other part of the code, of in another function. From the API perspective, this function can be called with a `ch` value that falls properly in a non-empty range, correctly expecting `Equal` to be return. Or did I misunderstand this?

unic/ucd/bidi/src/bidi_class.rs
```diff
@@ -215,6 +214,8 @@ impl TotalCharProperty for BidiClass {
 }
 
 
+// UCD/extracted/DerivedBidiClass.txt:
+// "All code points not explicitly listed for Bidi_Class have the value Left_To_Right (L)."
```
behnam: nit: Feel free to markdown up this and make doc block.

unic/utils/src/tables.rs
```diff
@@ -1,56 +1,89 @@
...
-    fn find_or(&self, needle: char, default: V) -> V {
-        self.find(needle).unwrap_or(default)
+impl<V> Default for CharDataTable<V> {
+    fn default() -> Self {
+        CharDataTable::Direct(&[])
     }
+}
 
-    /// Find the associated data for a character in this table or generate a default.
-    fn find_or_else<F>(&self, needle: char, f: F) -> V
```
behnam: Just wondering: How come decided to drop `find_or_else` and `find_or` and use a custom name?

CAD97: `find_or_else`/`find_or` are duplication of behavior from the `Option`. I think it's clearer to use the `Option` fn. That, and I'm not even really sure as to what the order of the parameters should be.

`find_defaulting` doesn't have any other call that clearly explains what's going on, and is a useful alias for `find_or_else(ch, Default::default)` (or `find(ch).unwrap_or_else(Default::default)` if we don't have `find_or_else`). That, and since I've bound it on `Copy` it makes working with these `Copy + Default` types a lot easier. The name might be better off as `find_or_default` though.

TL;DR: `find_or_else`/`find_or` don't add functionality over using `Option` fn, `find_or_default` (currently called `find_defaulting`) does.

unic/char/range/src/range.rs
CAD97: This specifically is talking about the empty-range case. How could I make this clearer?

unic/ucd/bidi/src/bidi_class.rs
CAD97: 👌

unic/utils/src/tables.rs
CAD97: That, and `find_or_else`/`find_or` (the concepts) aren't used in any of the code since all of our tables are called using `find_defaulting`.

unic/char/range/src/range.rs
behnam: Hum... but doesn't look like the language matches the code, for the case of release mode.

For example, if we have `chars!('z'..='a')`, this is an empty range, but comparing it to `'b'` will return `Ordering::Less` and not `Ordering::Equal`.

I think we need a more consistent behavior for empty ranges here. Since the signature doesn't allow to err, I guess it's better to just panic, in both debug and release.

What do you think?

unic/utils/src/tables.rs
behnam: Yeah, sounds good. Agree that `find_or_default()` would be a better name.

unic/char/range/src/range.rs
CAD97: It's definitely simpler to panic. In any case, since we panic in debug mode you have to avoid it anyway.

But I've intentionally not returned `Ordering::Equal` -- `::cmp(ch) == Ordering::Equal` should be equivalent to `::contains(ch)`. That's what the docs say -- that the ordering is never `Equal` for an empty range, but rather an arbitrary value which is not `Equal`. If that's not clear, the docs should be improved.

I haven't benchmarked it, but since we have to call this so much when binary searching through a table, I want to keep it `debug_assert!` if at all possible. But I think I'll move this to always panic for the time being; we can do that optimization later if needed. (I'll make a note of the possibility.)

CAD97:
@behnam that should address all of the points.

unic/char/range/src/range.rs
```diff
@@ -148,6 +149,25 @@ impl CharRange {
...
+    /// Determine the ordering of this range and a character.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the range is empty. This fn may be adjusted in the future to not panic
+    /// in optimized builds. Even if so, an empty range will never compare as `Ordering::Equal`.
+    ///
+    pub fn cmp(&self, ch: char) -> Ordering {
+        // possible optimization: only assert this in debug builds
+        assert!(!self.is_empty(), "Cannot compare empty range's ordering");
```
behnam: Awesome! We should consider this, but as long as we make sure the `release` mode behavior can't make more troubles, like resulting in an infinite loop if called on empty range.

Another possible solution would be to consider having separate concrete implementations for empty and non-empty ranges. (I think `std` had something like that.) The rational would be that we have much less range manipulation than we have lookups like this, therefore knowing in advance if the range is empty or not can save lots of unnecessary checks here and there.

What do you think?

behnam:
bors: r+

Issue #134, opened 2017-09-02
Improve unic-ucd-bidi and ucd-bidi


behnam:
We need to add more char properties to `unic-ucd-bidi` and use them in `ucd-bidi` to support more features like bidi mirroring of chars.

Properties:
* `Bidi_Control`
* `Bidi_Mirrored`
* `Bidi_Mirroring_Glyph`

PR #139, opened 2017-09-15
[ucd/bidi] Implement Bidi_Mirrored UCD property

behnam:
* [gen] Add `bidi_mirrored.rsv` generation.

* Add `BidiMirrored` type to `unic-ucd-bidi` component, as
`BinaryCharProperty`.

* Add `is_bidi_mirrored(char)` helper for common use-cases where only the
boolean value is needed. This is a pattern we can apply to all Binary
properties, like `is_alphabetic(char)` and `is_white_space(char)`.

* Add `CharBidiMirrored` and `StrBidiMirrored` traits to provide easy
access to property via `char` and `str` types.

Tracker: <https://github.com/behnam/rust-unic/issues/134>


unic/ucd/bidi/src/bidi_mirrored.rs
```diff
@@ -0,0 +1,213 @@
...
+}
+
+
+use unic_char_property::{BinaryCharProperty, CharProperty, TotalCharProperty};
+
+
+/// Represents values of the Unicode character property
+/// [*Bidi_Mirrored*](http://www.unicode.org/reports/tr44/#Bidi_Mirrored).
+///
+/// The value is `ture` if the character is a "mirrored" character in bidirectional text, `false`
```
CAD97: s/ture/true

unic/ucd/bidi/src/bidi_mirrored.rs
```diff
@@ -0,0 +1,213 @@
...
+    }
+}
+
+
+impl BidiMirrored {
+    /// Find the character *BidiMirrored* property value.
+    pub fn of(ch: char) -> Self {
+        use unic_utils::CharDataTable;
+        BidiMirrored(
+            CharDataTable::<()>::find(&data::BIDI_MIRRORED_TABLE, ch).is_some(),
```
CAD97: This _shouldn't_ be required, you should be able to just write `data::BIDI_MIRRORED_TABLE.contains(ch)`. If you can't, I messed something up and need to fix it with the new data table structure.

unic/ucd/bidi/src/bidi_mirrored.rs
```diff
@@ -0,0 +1,213 @@
...
+impl Default for BidiMirrored {
+    #[inline]
+    fn default() -> Self {
+        BidiMirrored(false)
+    }
+}
+
+
+// NOTE: This cannot be generalized at the moment and needs to be implemented for concrete types
+// indivisually. See <https://users.rust-lang.org/t/12884>.
```
CAD97: s/indivisually/individually

behnam:
Thanks, @CAD97, for catching those! Yes, `contains()` works smoothly.

bors: r+

Issue #127, opened 2017-08-17
Mapping the UCD into unic-ucd subcrates


CAD97:
[UAX44 § 5.1 Property Index][UAX44§5.1] gives a list of UCD properties. For convenience, I have reproduced below those which are intended for exposure in library APIs. This issue will serve as a tracking list for exposing those. Each property is also given a type in [UAX44 § 5.3 Property Definitions][UAX44§5.3] (one of Catalog, Enumeration, Binary, String, Numeric, or Miscellaneous). For definitions of those, see [UAX44 § 5.2][UAX44§5.2]. The type is also included in the below table for ease of reference.

<details><summary><strong>Property Index</strong></summary>

## General
- [ ] Name (Miscellaneous)
- [ ] Name_Alias (Miscellaneous)
- [ ] Block (Catalog)
- [X] Age (Catalog)
- [X] General_Category (Enumeration)
- [ ] Script (Catalog)
- [ ] Script_Extensions (Miscellaneous)
- [ ] White_Space (Binary)
- [ ] Alphabetic (Binary)
- [ ] Hangul_Syllable_Type (Enumeration)
- [ ] Noncharacter_Code_Point (Binary)
- [ ] Default_Ignorable_Code_Point (Binary)
- [ ] Deprecated (Binary)
- [ ] Logical_Order_Exception (Binary)
- [ ] Variation_Selector (Binary)
## Case
- [ ] Uppercase (Binary)
- [ ] Lowercase (Binary)
- [ ] Lowercase_Mapping (String)
- [ ] Titlecase_Mapping (String)
- [ ] Uppercase_Mapping (String)
- [ ] Case_Folding (String)
- [ ] Simple_Lowercase_Mapping (String)
- [ ] Simple_Titlecase_Mapping (String)
- [ ] Simple_Uppercase_Mapping (String)
- [ ] Simple_Case_Folding (String)
- [ ] Soft_Dotted (Binary)
- [ ] Cased (Binary)
- [ ] Case_Ignorable (Binary)
- [ ] Changes_When_Lowercased (Binary)
- [ ] Changes_When_Uppercased (Binary)
- [ ] Changes_When_Titlecased (Binary)
- [ ] Changes_When_Casefolded (Binary)
- [ ] Changes_When_Casemapped (Binary)
## Numeric
- [ ] Numeric_Value (Numeric)
- [ ] Numeric_Type (Enumeration)
- [ ] Hex_Digit (Binary)
[...]

PR #141, opened 2017-09-15
[data/ucd] Add PropList.txt

behnam:
* [gen] Set package version to `0.0.0`


behnam:
bors: r+

CAD97:
I apparently had the same idea; I've already parsed in `DerivedCoreProperties` and getting in `PropList` will be mostly the same codepath; let me do it so we don't duplicate work 😛

behnam:
Uh, yeah, I had no idea you're working on it. Should have filed an issue first, so we can coordinate. I think we can keep that discussion in https://github.com/behnam/rust-unic/issues/127 now.

Issue #133, opened 2017-09-02
Implement unic-ucd-ident


behnam:
Char Properties:
* `ID_Start`/`ID_Continue`
* `XID_Start`/`XID_Continue`
* `Pattern_Syntax`/`Pattern_White_Space`

References:
* http://www.unicode.org/reports/tr44/tr44-4.html#ID_Start
* http://www.unicode.org/reports/tr31/

Cross-component tests:
* http://www.unicode.org/reports/tr31/#Table_Lexical_Classes_for_Identifiers

Similar crates:
* https://github.com/unicode-rs/unicode-xid

CAD97:
The main use-case I have for this crate has a want for the `XID_*` and `Pattern_*` properties. I therefor can knock this component out, but I'd like to land #136 first.

CAD97:
Request for a second opinion.

https://github.com/CAD97/rust-unic/commit/6969236a703bce87602c0286e614f1ec0f1997ef

This lightens the `*_continue` tables by a little under 1.5KB (1474 `char`) if my math is correct. However, it also means that checking a character's membership in `xid_continue` or `id_continue` will take about eight more comparisons (to a range) (over two tables instead of one). This does seem like a small difference once I did the math (`log2(line_count(new_continue)) - log2(line_count(old_continue)) + log2(line_count(start))`, but I want your opinion on whether the size optimization is worth it.

CAD97:
On the above:

I realized that this would mean that the `*_continue` properties could not use the macro constructor if I do that, as the `is_*_continue` fn would have to be modified to check both tables. I guess that's a path for future possible benchmark+optimization, but for now I'm going to not use that optimization.

PR #142, opened 2017-09-15
Gen useful shared files

CAD97:


CAD97:
So that's how close I was to submitting the pull request @behnam; really interesting that we both had the same idea. Of course, both #133 and #134 rely on PropList, so I guess it was inevitable.

CAD97:
bors: r+

That would be now. Just had to add the doc comments because I gotta stay in the habit of including them 😆

Issue #122, opened 2017-08-14
Make `EnumeratedCharProperty`'s `FromStr` impl fully UAX44-LM3 compliant


CAD97:
Here's the current implementation:

<https://github.com/behnam/rust-unic/blob/99164db5dd8f909cd5b491dcd516b13b3356c6ea/unic/char/property/src/macros.rs#L111-L129>

It does a quick check for simple equivalence against the abbreviated or long aliases, then does a case insensitive compare. However, for proper compliance we should comply with [UAX44-LM3].

> **5.9.3 [Matching Symbolic Values][UAX44-LM3]**
> 
> Property aliases and property value aliases are symbolic values. When comparing them, use loose matching rule UAX44-LM3.
> 
> *[UAX44-LM3]*. Ignore case, whitespace, underscore (`_`), hyphens, and any initial prefix string "is".
> 
> * "linebreak" is equivalent to "Line_Break" or "Line-break"
> * "lb=BA" is equivalent to "lb=ba" or "LB=BA"
> * "Script=Greek" is equivalent to "Script=isGreek" or "Script=Is_Greek"
> 
> Loose matching is generally appropriate for the property values of [our `EnumeratedCharacterProperty`], which have symbolic aliases defined for their values. Loose matching should not be done for the property values of String properties, which do not have symbolic aliases defined for their values; exact matching for String property values is important, as case distinctions or other distinctions in those values may be significant.
> 
> For loose matching of symbolic values, an initial prefix string "is" is ignored. The reason for this is that APIs returning property values are often named using the convention of prefixing "is" (or "Is" or "Is_", and so forth) to a property value. Ignoring any initial "is" on a symbolic value during loose matching is likely to produce the best results in application areas such as regex. Removal of an initial "is" string for a loose matching comparison only needs to be done once for a symbolic value, and need not be tested recursively. There are no property aliases or property value aliases of the form "isisisisistooconvoluted" defined just to test implementation edge cases.
> 
[...]

PR #143, opened 2017-09-15
Add experimental binary property support to char_property!

CAD97:
This should not be merged yet; I would like to see a bit more example implementations of binary properties before pulling in a version of binary property support in the `char_property!` macro.

But a preliminary "how's the syntax & doc look" would be useful.


CAD97:
That now reduces the duplication I wanted to factor out and fleshes out the test of the new macro usage. I still think it's probably better to get a bit more use-time with `BinaryCharacterProperty` before merging the macro, but I'll leave it to your digression @behnam.

CAD97:
Note to self that this effects #122, changing the current FromStr handling. It should make the task simpler.

CAD97:
Oh fun. A 1.17 failure...

```
error: expected pattern, found `stringify!(AL)`
  --> <char_property macros>:66:5
   |
66 | $ ( $ repr => Ok ( $ value ) , ) * $ (
   |     ^^^^^^
```

I'll get to that tomorrow.

CAD97:
I suspect the macro resolution change that allows `$(#[$meta])* $ident` to work in 1.20 is related to why the above failure happens. b159cf6 contains a simple workaround -- move the `stringify!` call inline so it is expanded properly, but should probably be reverted when the minimum compiler version is bumped to 1.20, along with the other 1.20 related macro restrictions in place. I've doc'd it inline.

unic/char/property/src/macros.rs
```diff
@@ -48,13 +48,38 @@
...
+///     /// This is the newtype used for the character property.
+///     pub struct MyProp(bool) {
+///         abbr => "AbbrPropName";
+///         long => "Long_Property_Name";
+///         human => "Human-Readable Property Name";
+///     }
+///
+///     /// Unlike an enumerated property, a binary property will handle the table for you.
+///     /// This requires `unic_utils` to be in scope.
+///     mod data = "path/to/table.rsv";
```
behnam: Since the `data` module is not exposed (and in this case, since it's auto-created, we can actually skip it), I think this one is better as something like:
```
data_table => "path/to/table.rsv";
```
similar to setting name strings.

What do you think?

unic/char/property/src/macros.rs
```diff
@@ -108,62 +136,122 @@ macro_rules! char_property {
...
+
+        $(#[$data_mod_meta])*
+        mod $data_mod {
+            use super::unic_utils::CharDataTable;
+            pub const TABLE: CharDataTable<()> = include!($data_path);
+        }
+
+        impl $name {
+            pub fn of(ch: char) -> Self { $name(data::TABLE.contains(ch)) }
+            pub fn bool(&self) -> bool { self.0 }
```
behnam: Looks like this one, `bool()`, actually is better get a verb-ish name, like `is_mirrored` for Bidi_Mirrored property, or `changes_when_lowercased()` for `Changes_when_Lowercased` prop. We didn't have this for Enumerated types and it's Binary-specific.

What do you think about this? Should we take a name in input and use it here?

unic/char/property/src/macros.rs
```diff
@@ -108,62 +136,122 @@ macro_rules! char_property {
...
+            __impl FromStr for $name;
+            y => $name(true); yes => $name(true); t => $name(true); true => $name(true);
+            n => $name(false); no => $name(false); f => $name(false); false => $name(false);
+        }
+
+        char_property! { __impl CharProperty for $name; $abbr_name; $long_name; $human_name; }
+        char_property! { __impl Display for $name by BinaryCharProperty }
+
+        impl $crate::BinaryCharProperty for $name {
+            fn bool(&self) -> bool { self.bool() }
```
behnam: And then this one needs to be updated to use the shorthand name.

unic/char/property/src/macros.rs
```diff
@@ -108,62 +136,122 @@ macro_rules! char_property {
...
+            pub const TABLE: CharDataTable<()> = include!($data_path);
+        }
+
+        impl $name {
+            pub fn of(ch: char) -> Self { $name(data::TABLE.contains(ch)) }
+            pub fn bool(&self) -> bool { self.0 }
+        }
+
+        impl From<bool> for $name {
+            fn from(b: bool) -> Self { $name(b) }
```
behnam: Actually, I don't think we want this one. We prefer to prevent creating values manually in general, right? I prefer we provide const values, like YES and NO, instead of conversion from boolean. What do you think?

CAD97:
Closed in favor of #154.

PR #145, opened 2017-09-15
[gen/download] Simplify the config format and download script

behnam:
There's no more code in the download module for each source, and the
list of source comes only from the config file, as well as the file
paths.

For each source, two types of data is supported: normal (or main)
data, and test data. The former goes into `/data/<source>/<name>`, while
the latter goes into `/data/<source>/test/<name>`.

The config file is also simplified to not have any duplicate content as
much as possible.

One major pattern to follow is that each data source has a *base URL*,
which contains the domain name and path to a specific directory for the
version. Nothing goes wrong in the code if this is not maintained, but
it's good practice to keep it so, and create new source config tables if
the domain or path-to-version are different.


gen/src/download/mod.rs
```diff
@@ -8,107 +8,86 @@
...
+//!
+//! ```toml
+//! [example]
+//! version = "10.0.0"
+//! base_url = "http://example.com/path/to/{version}/"
+//!
+//! [[example.resources]]
+//! "some_data.txt" = "some_data.txt"
+//!
+//! [[example.test.resources]]
```
CAD97: This doesn't agree with the file format actually currently in use.

In use:
```toml
[example]
[example.resources]
[example.test_resources]
```

Shown here:
```toml
[example]
[[example.resources]]
[[example.test.resources]]
```

gen/src/download/mod.rs
```diff
@@ -8,107 +8,86 @@
...
-pub fn download(components: &[&str]) -> Result<(), Box<Error>> {
-    let mut file = File::open(Path::new(DOWNLOADS)).expect("Failed to open config.toml");
-    let mut buffer = String::new();
-    file.read_to_string(&mut buffer)
-        .expect("Failed to read config.toml");
-
-    let config: Config = toml::from_str(&buffer).expect("Failed to parse downloads.yaml");
+pub fn download(active_sources: &Vec<&str>) -> Result<(), Box<Error>> {
+    let config: Config =
+        toml::from_str(include_str!("config.toml")).expect("Failed to parse config file");
```
CAD97: Note: Using `include_str!` here means recompiling the module/crate when the config changes. That's the reason I originally used the file APIs to get at it, such that we could add more downloads without having to recompile.

It doesn't add much complexity to read the config file in at runtime rather than compile time, so I personally think its worth it.

CAD97: But I'm willing to go either way. Updating the config should be an infrequent enough thing that this doesn't matter.

gen/src/download/mod.rs
behnam: Uh, good catch. I forgot to update it here.

gen/src/download/mod.rs
behnam: I think having less lines of code is a good goal here. Since running this tool is only useful in a dev environment, I don't see any benefit of not doing the simplest and shortest option. So, considering this and your comment, let's keep it then.

behnam:
> It seems a little silly to do a replace for version when the version is specified literally one line above the string it's going to live in. But I suppose that is nice for being able to check at a glance (and update easily) the version of the data store we are checking.

True. I think the updates (supposedly) only affecting the version number is a good-enough reason to keep it separate.

Fixed the typo in the doc. Should be good to go.

gen/src/download/mod.rs
```diff
@@ -8,107 +8,86 @@
...
+//!
+//! ```toml
+//! [example]
+//! version = "10.0.0"
+//! base_url = "http://example.com/path/to/{version}/"
+//!
+//! [[example.resources]]
+//! "some_data.txt" = "some_data.txt"
+//!
+//! [[example.test_resources]]
```
CAD97: You changed the `.` to a `_` but missed the fact that you've double bracketed the headers here incorrectly.

Maybe that's not as obvious a syntax as TOM thinks it is 😛

CAD97: (disclaimer: that was a joke by a tired brain)

behnam: 🤣 Yeah, that one is **really** hard to see!!!

PR #146, opened 2017-09-15
Move tables dirs out of src

behnam:
All files under `tables` directories are auto-generated and should not
be affected by tools working on manually written code, like linters.

We need to make sure, though, that we package the `tables` directories,
otherwise the source won't compile.


behnam:
Thanks for the quick review.

Yeah, unfortunately cargo doesn't have tools to validate built packages and it's something that has to be done manually. I'm fine with either `include` or `exclude`. The only thing is that `include` will be more verbose. I'm not against changing it, though.

bors: r+

PR #147, opened 2017-09-17
[gen] Split generate module into reader and writer

behnam:
Generating data files useful for Rust code from data sources is not
always a one-to-one mapping. First, we have the fact that some sources,
like `UnicodeData.txt` provides data for many different UNIC components.
Second, we have components that depend on multiple sources. We don't
want to duplicate any code for processing the data, and don't want to
make random cross-file dependencies. So, it's better to split the
modules into two parts: reading source files, and write into component
table files.

The `reader` module provides access to raw data files (from `/data/`)
and is only responsible of parsing the file and returning the value(s)
we're interested in. For example, `reader::ucd::readme` reads
`/data/ucd/ReadMe.txt` file and provides `UNICODE_VERSION`, which is
extracted data from the file content. Most files only provide one big
data table.

All data reading in `reader` is performed lazily, via `lazy_static`,
freeing caller from loading data files explictly and passing around
instances.

The `writer` module is responsible for generating `tables` files for
`unic` components. For example, `writer::ucd::age` is responsible for
creating files under `/unic/ucd/age/tables/`, which depends on
`reader::ucd::readme` (for `UNICODE_VERSION`) and
`reader::ucd::derived_age` (for `AGE_DATA`).

Other Notes:

We were using `std::io::Result` for our error-management, which is not
flexible enough for this case, as different kinds of `Error`s need to be
supported. Because it was a half-baked solution, I just dropped all of
it and replaced error-handling with `unwrap()` calls. We can (and
should) fix this later by using `error-chain` crate and define a clear
error-handling process. Until then, `unwrap()` and `expect()` work good
enough.

Also simplified many fn signatures with dropping the generics and switch
to concrete types, like `PathBuf`, which is good enough for a scripting
enviroment, when we know that we're only passing in one type of value.
We can change this in the future, but only if it really is needed.
Otherwise, it's just extra complexity for a simple task.


behnam:
What do you think about this approach, @CAD97? I think it's a good balance for keeping source and components separate, without extra overhead.

Btw, another step I want to take before landing this is to replace the iter-based `from()` implementations to a simpler one, using source data internally and returning the data structure we're interested in.

My goal is to make it faster and easier to add new sources and components, and this is along that path.

behnam:
And here's the current output of the gen script:
```
$ c run --package=unic-gen -- --generate ucd normal idna
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target/debug/unic-gen --generate ucd normal idna`

    Clean up dir: = unic/idna/mapping/tables
    Reading file: < data/idna/ReadMe.txt
    Writing file: > unic/idna/mapping/tables/unicode_version.rsv
    Reading file: < data/idna/IdnaMappingTable.txt
    Writing file: > unic/idna/mapping/tables/idna_mapping.rsv

    Clean up dir: = unic/ucd/core/tables
    Reading file: < data/ucd/ReadMe.txt
    Writing file: > unic/ucd/core/tables/unicode_version.rsv

    Clean up dir: = unic/ucd/age/tables
    Writing file: > unic/ucd/age/tables/unicode_version.rsv
    Reading file: < data/ucd/DerivedAge.txt
    Writing file: > unic/ucd/age/tables/age_values.rsv

    Clean up dir: = unic/ucd/bidi/tables
    Writing file: > unic/ucd/bidi/tables/unicode_version.rsv
    Reading file: < data/ucd/UnicodeData.txt
    Writing file: > unic/ucd/bidi/tables/bidi_class_values.rsv
    Writing file: > unic/ucd/bidi/tables/bidi_mirrored.rsv

    Clean up dir: = unic/ucd/category/tables
    Writing file: > unic/ucd/category/tables/unicode_version.rsv
    Writing file: > unic/ucd/category/tables/general_category.rsv

    Clean up dir: = unic/ucd/name/tables
    Writing file: > unic/ucd/name/tables/unicode_version.rsv
    Writing file: > unic/ucd/name/tables/name_values.rs
    Writing file: > unic/ucd/name/tables/name_map.rsv

    Clean up dir: = unic/ucd/normal/tables
    Writing file: > unic/ucd/normal/tables/unicode_version.rsv
    Writing file: > unic/ucd/normal/tables/general_category_mark.rsv
    Writing file: > unic/ucd/normal/tables/canonical_combining_class_values.rsv
    Writing file: > unic/ucd/normal/tables/canonical_decomposition_mapping.rsv
    Reading file: < data/ucd/DerivedNormalizationProps.txt
    Writing file: > unic/ucd/normal/tables/canonical_composition_mapping.rsv
    Writing file: > unic/ucd/normal/tables/compatibility_decomposition_mapping.rsv

    Clean up dir: = unic/normal/tests/tables
    Reading file: < data/ucd/test/NormalizationTest.txt
    Writing file: > unic/normal/tests/tables/conformance_tests_data.rsv
```

gen/src/reader/utils.rs
```diff
@@ -0,0 +1,24 @@
...
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+
+use std::fs::File;
+use std::io::Read;
+use std::path::PathBuf;
+
+
+// TODO: Get source ID and call download for file path?
```
CAD97: Unless we want to parallelize the generation as well, doing download and generate in two separate steps is probably the best idea. The parallelization of downloads that Hyper allows to happen is why the downloads go almost sillily quickly.

gen/src/writer/ucd/bidi.rs
```diff
@@ -27,16 +27,10 @@ use generate::tables::{ToRangeCharSet, ToRangeCharTable};
 struct BidiClassData<'a>(BTreeMap<char, &'a str>);
 
 impl<'a> BidiClassData<'a> {
-    fn emit<P: AsRef<Path>>(&self, dir: P) -> io::Result<()> {
+    fn emit(&self, dir: &PathBuf) {
```
CAD97: If we're going to make these concrete types, we should use `&Path` instead of `&PathBuf`. `Path` is the slice type, and `&PathBuf` coerces to `&Path`.

gen/src/reader/ucd/unicode_data.rs
```diff
@@ -175,15 +182,10 @@ pub struct UnicodeDataEntry {
     // by unsafe-hacking a u32::MAX char, and not losing any application space.
 }
 
+
 #[derive(Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
-pub struct UnicodeData(Box<[UnicodeDataEntry]>);
+pub struct UnicodeData(pub Box<[UnicodeDataEntry]>);
 
-impl UnicodeData {
-    pub fn iter(&self) -> Iter<UnicodeDataEntry> {
```
CAD97: Why did you remove this fn? It's much clearer at the use site to call `UNICODE_DATA.iter()` rather than `UNICODE_DATA.0.iter()`.

I say use the following:

```rust
pub struct UnicodeData(Box<[UnicodeDataEntry]>);
impl UnicodeData {
    pub fn iter(&self) -> Iter<UnicodeDataEntry> {
        self.0.iter()
    }
}
```

It's just much cleaner not to have to `.0.iter()` at the use site.

CAD97: Of course, then you can bring up the other ones that wrap a `To*DataTable`. Honestly, I don't know. You could to a pass-through impl of that trait there as well, but that unnecessary code just shouldn't be required in this script environment. So I don't really have an opinion on this one, I've decided.

CAD97:
Since we're this deep into the generation code at this point:

How about we consider downloading + generating the tables at test time? That way they don't have to be in the repo, and there's (less) danger of the generated files accidentally not being updated and getting out of sync with the generating code, or using the wrong data source files. It _would_ add another step before someone could start hacking on the code, and would increase CI times, but I think it would be a net gain for the repository. It's a general guideline that you shouldn't be checking in trivially regenerable files.

gen/src/reader/utils.rs
behnam: This one was just about not hardcoding paths and instead calling the `download` code to give us the path, having the source name, test status, and file name. But I don't think we're going down that path, as keeping the download script completely separate looks the best, for reasons I'm going to explain in a follow-up comment.

gen/src/writer/ucd/bidi.rs
behnam: Right. I'll fix them.

gen/src/reader/ucd/unicode_data.rs
behnam: For all the data models under `reader`, I've made the internals public, so we don't have to write and maintain wrappers. I'll explain more in the follow up.

behnam:
> How about we consider downloading + generating the tables at test time? That way they don't have to be in the repo, and there's (less) danger of the generated files accidentally not being updated and getting out of sync with the generating code, or using the wrong data source files. It would add another step before someone could start hacking on the code, and would increase CI times, but I think it would be a net gain for the repository. It's a general guideline that you shouldn't be checking in trivially regenerable files.

So, here's the ideas and requirements I have in mind:

1. There's no reliable way to access Unicode data files at the moment. Using HTTP to download the files from a server that can go down or block IP addresses easily, is a bottle-neck for having any part of data source automated, without supervision. That's why we are keeping a copy of the files in the project since the beginning, and won't be able to change it until the upstream issue is addressed. (I'm following this up with Unicode.)

2. Because of (1), we can't have unsupervised downloads for the source data. And since we have the source data included, I don't see any benefit in spending lots of time to keep the table-generation code working on different environments, and it's much easier if we look at that part of the code as a developer-only code that only needs to be run on developers' systems. Also, again, there's no point in keeping this *script-style* code generic, having a nice API. It just needs to be able to generate the table files fast, without failure.

3. We have a lot of new components to implement and what I'm trying to do here is to simplify the table-generation step for these components. I think we really shouldn't spend much more time in this area. In fact, I think it's totally fine now to keep the download and generation parts coupled, as it's not an issue anymore and it's flexible and simple enough.

4. I think having all the data models for source files in the memory (lazy-static) is just fine, as our (developers) machines can easily handle this amount of data and I don't think UNIC can grow that much that our >8 GB machines can go out of memory because of the loading some source data.

Of course, we can always improve any of these areas in the future when they become a problem. I think we just don't need to forecast any issues like this in the developer-only code, and we can just keep the goal as minimum-lines-of-code.

What do you think?

behnam:
Okay, I have reorganized and renamed most of the gen modules/functions to map their external structure (source or unic components).

Also, I have converted all writer functions to be functional, and putting the shared intermediate data into their own `lazy-static`ed functions. This way, we don't make much abstractions by type, but by only the data processing functionality, which is easier to read and follow, at a cost of some possible extra `clone()`s.

I think we're good to go here now.

bors: r+

behnam:
bors: r+

PR #148, opened 2017-09-17
[gitattributes] Update pattern for RSV files

behnam:


CAD97:
blargh the RenderScript is back...

PR #151, opened 2017-09-18
[ucd/bidi] Implement Bidi_Class property

behnam:
* Uses PropList data.

* [gen] Convert remaining tuple structs to named.

* [char/property] Add note about range type enforcement.


behnam:
bors: r+

PR #152, opened 2017-09-19
[unicode_version_tests] Re-enable std::char comparison

behnam:
Also, add TODO items for adding convertors to/from internal
`UnicodeVersion` type.


behnam:
bors: r+

PR #154, opened 2017-09-19
Add binary type support to char_property macro

behnam:
This is an update to <https://github.com/behnam/rust-unic/pull/143>, expanding the macro syntax, and using the macro for existing binary UCD properties.


unic/char/property/src/macros.rs
```diff
@@ -108,63 +147,229 @@ macro_rules! char_property {
...
             fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
-                $crate::EnumeratedCharProperty::human_name(self).fmt(f)
+                $crate::$trait::human_name(self).fmt(f)
+            }
+        }
+    };
+
+    ( __impl CharTraitBinary for $name:ident;
+        $bool_name:ident;
+        $(#[$char_trait_meta:meta])* $char_trait_name:ident;
```
CAD97: Here's the 1.17 failure.

```
  --> unic/ucd/bidi/src/bidi_control.rs:19:1
   |
19 | char_property! {
   | ^
   |
   = note: this error originates in a macro outside of the current crate
```

You can't have a repeating group terminated by an `:ident` token until 1.20. You'll need to add an explicit sigil between the meta and the ident, or move the meta after the ident.

I'll do a more full review later today, but I wanted to point this out quickly so you'd know what to fix.

behnam: Thanks, @CAD97! Fixed in the a new commit.

unic/char/property/src/macros.rs
```diff
@@ -108,62 +147,220 @@ macro_rules! char_property {
...
+        pub fn $shorthand_name:ident(char) -> bool;
+
+        $(#[$char_trait_meta:meta])*
+        pub trait $char_trait_name:ident for char {
+            $(#[$char_trait_fn_meta:meta])*
+            pub fn $char_trait_fn_name:ident(self) -> bool;
+        }
+
+        $(#[$str_trait_meta:meta])*
+        pub trait $str_trait_name:ident for str {
```
CAD97: Does it make sense for _all_ binary properties to have a trait for `str`? And if we do, wouldn't it make sense just to reuse one trait for both? (So a (strawman) `UnicodeLowercase` trait that offers the fn `is_unicode_lowercase`, which is impl'd for both `char` and `str`.)

Some cases will definitely want a `str` impl. But I think some wouldn't: does `string.is_question_marks` seem like useful functionality, or something that's better left out of the API and just expressed as the marginally more complicated `string.chars().all(|c| c.is_question_mark())`?

unic/char/property/src/macros.rs
```diff
@@ -108,62 +147,220 @@ macro_rules! char_property {
...
+            abbr => $abbr_name:expr;
+            long => $long_name:expr;
+            human => $human_name:expr;
+
+            data_table_path => $data_path:expr;
+
+            pub fn $bool_name:ident(&self) -> bool;
+        }
+
+        pub fn $shorthand_name:ident(char) -> bool;
```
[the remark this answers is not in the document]
CAD97: But also, what's the point of having the freestanding fn when the trait exists? It seems like two ways of getting at the same thing, and the trait fn is overloadable to take both characters and strings as well.

unic/char/property/src/macros.rs
```diff
@@ -108,62 +147,220 @@ macro_rules! char_property {
...
+
+        $(#[$char_trait_meta:meta])*
+        pub trait $char_trait_name:ident for char {
+            $(#[$char_trait_fn_meta:meta])*
+            pub fn $char_trait_fn_name:ident(self) -> bool;
```
CAD97: If we keep this trait, it should borrow.

unic/char/property/src/macros.rs
```diff
@@ -108,62 +147,220 @@ macro_rules! char_property {
...
+
+        impl $crate::TotalCharProperty for $name {
+            fn of(ch: char) -> Self { Self::of(ch) }
+        }
+
+        impl $crate::BinaryCharProperty for $name {
+            fn bool(&self) -> bool { self.$bool_name() }
+        }
+
+        impl From<bool> for $name {
```
CAD97: You mentioned dropping this impl in #143. Do you still want to?

I actually don't see much value in e.g. `UnicodeLowercase::from(true)` anymore. What would it even be used for? It can always be re-added later. (Or, similarly, ever using an `Into<UnicodeLowercase>`.)

unic/char/property/src/macros.rs
```diff
@@ -48,13 +48,44 @@
...
+///     // A function that returns whether the given character has the property or not.
+///     pub fn is_property(char) -> bool;
+/// }
+/// ```
+///
 /// # Effect
 ///
 /// - Implements the `CharProperty` trait and appropriate range trait
 /// - Implements `FromStr` accepting either the abbr or long name, ascii case insensitive
+/// - Implements `From` to/from the wrapped type
```
CAD97: I've called whether we actually want to do this for binary properties into question below. Make sure to update this along with the decision we make.

unic/char/property/src/macros.rs
```diff
@@ -108,62 +147,220 @@ macro_rules! char_property {
...
+
+        impl $crate::BinaryCharProperty for $name {
+            fn bool(&self) -> bool { self.$bool_name() }
+        }
+
+        impl From<bool> for $name {
+            fn from(b: bool) -> Self { $name(b) }
+        }
+
+        impl From<$name> for bool {
```
CAD97: Similarly to above, do we actually want this? We already have a fn to turn a property into a bool anyway. Is there any use case for using `Into<bool>`? Otherwise I feel that this impl is actually just bloat rather than useful. It could be added later if we find a use for it.

unic/char/property/src/macros.rs
```diff
@@ -108,62 +147,220 @@ macro_rules! char_property {
...
+    ) => {
+        $(#[$str_trait_meta])*
+        pub trait $str_trait_name {
+            $(#[$str_trait_fn_meta])*
+            fn $str_trait_fn_name(&self) -> bool;
+        }
+        impl $str_trait_name for str {
+            #[inline]
+            fn $str_trait_fn_name(&self) -> bool {
+                self.chars().any(|ch| $name::of(ch).$bool_name())
```
CAD97: I would have thought that a str impl would check `all`, not `any`.

CAD97: I can see where the has-a definition is convenient for the bidi properties. But unfortunately, I don't think that applies to all properties.

unic/char/property/src/macros.rs
```diff
@@ -108,62 +147,220 @@ macro_rules! char_property {
...
+
+            data_table_path => $data_path:expr;
+
+            pub fn $bool_name:ident(&self) -> bool;
+        }
+
+        pub fn $shorthand_name:ident(char) -> bool;
+
+        $(#[$char_trait_meta:meta])*
+        pub trait $char_trait_name:ident for char {
```
CAD97: I wonder if this trait is pulling its weight. you don't use it in the implementation of `CharBidiClass`, and I suspect most, _if not all_ of the UCD crates will want to expose a trait that has fn for all of its exposed properties. That's just much more convenient to use than a trait for _every single property_. Would you rather:

```rust
import unic_ucd_casing::{UnicodeLowercase, UnicodeUppercase, UnicodeTitlecase, ... };
// or
import unic_ucd_casing::UnicodeCasing;
```

(In the casing case, I suspect we want one trait which exposes a version of `is_lowercase` and friends, and works for both `char` and `str` (in the obvious all form, not the has-a form).)

unic/char/property/src/macros.rs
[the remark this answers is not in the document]
CAD97: I figured out where the shorthand name comes from. After a bit of thinking, I think I'm in favor of keeping this fn, and dropping the in-macro traits.

CAD97:
I'm making a patch right now based off of my above feedback. Hopefully we can find a design we agree on 🙃

behnam:
Thanks, @CAD97, for the feedback. Agree that we better keep the macro minimal. Based on you #155 PR, I addressed some small issues and updated this branch to pass all the tests. Plus some small fixes here and there.

Let's land this minimal version, and we can refine more if you have any more feedback. I have another branch with some properties that I like to land before we make small changes to the macro. I think that can help with improving the design, that's why I think it's better to have this in asap.

behnam:
bors: r+

CAD97:
Looks good to me! 👍

PR #155, opened 2017-09-20
char_property binary property support

CAD97:
Update to #154. See #154's review before the creation of this PR for a lot of my reasoning.


CAD97:
Oh right, there was a change sometime between 1.17 and now that changed the behavior of `include!` in doctests. In current Rust, it's relative to where it's written. In 1.17, I believe it's relative to the root of the project.

Ugh. The fix is to reignore the binary property doctest, but that just makes me sad.

See https://github.com/rust-lang/rust/issues/34431

behnam:
So, @CAD97, should we land this first, and I rebase the left-over matters on top of this? Want to fix the error, then?

behnam:
By the way, about the `tables` dir, for when it's part of the package (library) source, it goes under `tables/` in package root. When it's only used for tests, it goes under `tests/tables/`. This is similar to how we organize the source data, just to make it clear what data is used in library and what's used in dev/test.

This came up about in another PR and I forgot to answer there.

And, here, `unic/char/property/tables/property_table.rsv` should be in `unic/char/property/tests/tables/property_table.rsv`.

PR #156, opened 2017-09-20
[utils] Drop codepoints module

behnam:
* We have developed `unic-char-range` module to replace
`iter_all_chars()`, and there no more functionality left in
`unic-utils`'s `codepoints` module, therefore dropping the module
completely.

* Replace all `iter_all_chars()` calls with `CharRange::all()`.

* It's common practice to start the name of all test functions (those
with `#[test]` attr) with `test_`, to help differentiating them from
other functions. We follow it for most cases, but were missing it in
some.

* Add copyright notice to files missing it.


behnam:
bors: r+

PR #162, opened 2017-09-23
Add unic-ucd-ident

### LINKED ISSUE #133: Implement unic-ucd-ident

CAD97:
Closes #133 

The three pairs of properties are individual features on the idea that you probably only need one of them.


unic/ucd/ident/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+
+#[macro_use]
+extern crate unic_char_property;
+#[macro_use]
+extern crate unic_char_range;
+extern crate unic_ucd_core;
+
+
+#[cfg(feature = "xid")]
+mod xid {
```
behnam: This looks good when the module is this short. But would be great to expand over time with some basic unit tests, specially for ASCII, which is very relevant here. Not a blocker, though.

unic/ucd/ident/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+    char_property! {
+        /// A character that can start an identifier, stable under NFKC.
+        pub struct XidStart(bool) {
+            abbr => "XIDS";
+            long => "XID_Start";
+            human => "NFKC Identifier Start";
```
behnam: I think we better stick as close to the `long` format as possible, and only expand abbreviations, if there's any. Using "NFKC Identifier" for "XID" here is not common at all and I think we better keep it as simple as "XID Start".

Basically, how I look at the `human` value is "what we expect to see in a table header", where we don't want to stay machine-readable or have underscore joins, and there's room to not over-abbreviate.

unic/ucd/ident/src/lib.rs
CAD97: Since the tests would be for external API, it seems (to me) like they'd be better suited in the tests directory. But yeah, this can be addressed later.

unic/ucd/ident/src/lib.rs
CAD97: ID Start is obviously an "Identifier Start". XID Start is ID Start modified for closure under NFKC. Thus, "NFKC Identifier Start".

But yes, just using human => "XID Start" would be less confusing.

behnam:
bors: r+

Issue #140, opened 2017-09-15
Decouple download and generate scripts


behnam:
The requirements we have for download and generate scripts are fairly different.

For download, we want to keep files from every *source* in sync, as we read the version from their `ReadMe.txt`. So, for each source, we want to delete all the files before re-downloading. The files under `/data/`, for each source, should never go out of sync, version-wise. Because of this, this part of the code should be grouped together based on the data *source*.

For generate, in contrast, we want to have steps the generate consistent table files, which depend on one or more sources, and generates one or more table files. For this part of the code, it makes sense to group them based on *major component*, because that's closer to the development process.

The current model that has coupled download and generate code to be called from one binary makes it hard to satisfy this pattern.

Also, with the current model, we cannot add the download logic for a new source, like `emoji`, in one step and get to the generate code later; forcing us to put in a dummy generate code until we get to the implementation.

Also, the two steps don't share any model configuration, and only sharing some CLI implementation.

I think the current model doesn't have much benefits and it's going to be much easier if we have two separate creates for these tasks, each with their own functionality.

Also, now we can rename `gen` to `tools`, as we have killed all the Python code. So, we can have `/tools/download/` and `/tools/generate/`, each a separate binary create. Or, we can have one `tools` create with multiple binaries inside.

behnam:
What do you think about this, @CAD97?

CAD97:
I still think it's very useful to have a single command to both download and generate all of the necessary things. But that can be handled by a shell script, I suppose.

I'm fully open to decoupling the two. They'll always be somewhat coupled as `generate` relies on `download` having run and agreeing on where files are put, but it does make sense to split the two so they can have separate concepts of "data source" versus "component destination".

Maybe then we can actually get the cached-compile benefit for the download portion (the reason there's a config read in from the filesystem rather than just being hardcoded or `import!`ed) even while adding new output to the generate portion.

I'm actually poking around in it currently to add parsing for `PropList.txt` and `DerivedCoreProperties.txt`, which are both big shared files including many binary properties. Once I'm done with that, I can do this this weekend.

TL;DR I had my reasons for sticking them in one binary when building it -- mainly, being lazy -- but see only positives for splitting them now.

behnam:
Yeah, I think we started with the hope to share some data between them, like the data file paths. But, in practice, the path is hard-coded into an `include_str!` for the *generate* portion, while it's in a config file for the *download* portion, which demands more complexities just to share a few strings.

[...]

PR #163, opened 2017-09-26
Move data download code into unic-data crate

behnam:
* [`unic-data`] Create crate, hosted under `/data/`, which is responsible for downloading the files from sources, as set in `sources.toml`. This makes the `/data/` *component* standalone, by separating the matter of downloading data from the processing and table generation, which is the responsibility of `data-gen` crate.

* [`unic-gen`] Remove `--download` option and related code, and make `--generate` option the only behavior for the binary. Also make components list optional, making it easier to use the binary in other scripts, as adding new components won't need updates to other places.

* [`etc`] Add download-data.sh and rename `refresh-tables.sh` to `generate-tables.sh`.

Closes <https://github.com/behnam/rust-unic/issues/140>


behnam:
This solution also allows us to move `/data/` into its own git repository and use it here as a submodule, if we decide to do so.

data/src/config.rs
```diff
@@ -48,40 +30,42 @@ struct DataSource {
...
+/// Panics if the config file is badly formatted.
+pub fn get_download_maps(sources: &Vec<&str>) -> Vec<DownloadMap> {
+    let mut maps: Vec<DownloadMap> = vec![];
 
-    let mut downloads = vec![];
+    let config: Config = toml::from_str(CONFIG_TOML).expect("Failed to parse sources.toml file");
 
     for (source, info) in config {
-        if !active_sources.contains(&source.as_str()) {
+        if !sources.is_empty() && !sources.contains(&source.as_str()) {
```
CAD97: I had to go thorough a good deal of thinking to check if this was correct. I'm still not fully convinced. With only one condition, the continue is fine, but I think now it'd be better to reword this into the positive and wrap the following loops rather than have the continue.

behnam:
Thanks, @CAD97!

bors: r+

behnam:
bors: r+

PR #167, opened 2017-10-02
Use VecDeque for Decompositions.

clarcharr:
This optimises `remove(0)` to be `pop_front()` instead. This optimisation is used by the unicode-normalization crate but not this crate.

### DIFF
--- a/unic/normal/src/decompose.rs
+++ b/unic/normal/src/decompose.rs
@@ -11,13 +11,14 @@
 
 
 use std::fmt::{self, Write};
+use std::collections::VecDeque;
 
 use unic_ucd_normal::{decompose_canonical, decompose_compatible, CanonicalCombiningClass};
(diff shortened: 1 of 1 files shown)


unic/normal/src/decompose.rs
```diff
@@ -80,11 +81,11 @@ impl<I: Iterator<Item = char>> Iterator for Decompositions<I> {
         match self.buffer.first() {
```
CAD97: `Vec::first` -> `VecDeque::front`

CAD97:
Oh and also, thanks! A new pair of eyes spotting always helps out!

behnam:
Thanks, @clarcharr, for the PR. Looks great, just need to get all the checks green before we can land.

clarcharr:
It's passing now!

PR #168, opened 2017-10-02
Replace unnecessary display methods with to_string.

clarcharr:
The former implementation adds an unnecessary allocation to the `Display` impl. Because implementing `Display` gives the `to_string` method for free, there shouldn't be any `Display` methods.

If there are any methods which offer strings that differ from the `Display` implementation, those should have different names. I deprecated the methods instead of removing them, just to be safe.

### DIFF
--- a/unic/ucd/age/src/age.rs
+++ b/unic/ucd/age/src/age.rs
@@ -9,7 +9,6 @@
 // except according to those terms.
 
 
-use std::fmt;
 use std::cmp;
 
 pub use unic_ucd_core::UnicodeVersion;
@@ -89,13 +88,6 @@ impl Age {
     pub fn actual(&self) -> UnicodeVersion {
         self.0
     }
-
-    /// Human-readable description of the Age property value.
-    #[inline]
-    // TODO: Review return type
-    pub fn human(&self) -> String {
-        format!("Assigned in Unicode {}", self.0).to_owned()
-    }
 }
 
 
@@ -106,13 +98,6 @@ impl cmp::PartialOrd for Age {
             cmp::Ordering::Less => Some(cmp::Ordering::Greater),
             cmp::Ordering::Equal => Some(cmp::Ordering::Equal),
         }
-    }
-}
-
-
-impl fmt::Display for Age {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", self.human())
     }
 }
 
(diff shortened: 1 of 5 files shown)


unic/ucd/age/src/age.rs
```diff
@@ -92,9 +92,9 @@ impl Age {
 
     /// Human-readable description of the Age property value.
     #[inline]
-    // TODO: Review return type
+    #[deprecate(note = "use to_string method instead")]
```
CAD97: This should be `deprecated`, methinks

CAD97:
Correction: I was thinking of another fn when I said `human()` was a trait fn. I say get rid of the fake-ToString fn!

CAD97:
I can confirm that all tests pass just outright removing the fn you've marked `#[deprecated]` here. I'd r+ as soon as you made that change (or even just fixed the typo) but I want @behnam to weigh in here, since he's the one that added the `display` fn in the first place.

behnam:
Thanks, @clarcharr, for the suggestion and the PR. I'm generally not again this, and has been trying to keep the options open until we get more components and a better idea of the requirements. Since we have had some growth already, I think now is indeed a good time to clean this up.

The reason I've been insisting on keeping the `*human*()` methods around was for making it clear that this is only one of the possible ways to convert the value into string. But, since we have `impl Display` already, there's no need to duplicate it, I suppose.

But, if we want to get ride of `*human*()` methods, we should do that for all the places it's being used, and start calling them `display` again, as they are the default *display* form of the value.

So, it means we should update `char_property` macro and all `CharProperty` traits to use the name display for all cases, and update all implementations to have `display => ...` instead of `human => ...`.

This would make the behavior more clear to the user than what we have now, and that's a big plus, IMHO.

Also, since we have not locked down any parts of the API yet, there's no need to mark anything as deprecated. Just dropping old methods is fine at the moment.

What do you think, @clarcharr, @CAD97?

clarcharr:
The tests should be passing now!

One notable change: I removed the "Assigned in version" part from the `Display` for `Name` because it seems unintuitive, and I'm not sure if there'd ever be a context where someone would want that exact string and feel like it's too much of a burden to construct it themself with `format!`. Just printing the version makes more sense to me.

CAD97:
(Oh right, if we change `Age`'s `Display` impl you'll have to change the test that checks it as well)

clarcharr:
Makes sense!

In that case, I do think that it makes sense to just get rid of the `Display` impl for `Age` altogether, and leave it to just the `Debug` impl. I'll do that tomorrow when I get back to this.

CAD97:
I'm not the final say here, @behnam is. I just help out as much as I can, and spend way too much time right where I can respond almost instantly. Go with whatever he decides.

behnam:
Previously, I wanted the value display string fit in a sentence. But I think a better way to look at it is a nice table, under a column for the property. So, I think it's fine to delicate `Display` for `Age` to its `UnicodeVersion`. In case it needs any update, we can again do so later.

So, @clarcharr, all looks good here. Would you please the tests accordingly and we get it merged in?

clarcharr:
I have done the thing! How does it look?

behnam:
bors: r+

PR #170, opened 2017-10-02
Replace SentenceBreak::Other with Option::None

CAD97:
In https://github.com/behnam/rust-unic/pull/166#issue-261969328 you said

> Current implementation uses `char_property!()` macro for
`EnumeratedCharProperty` implementation, which only supports
`TotalCharProperty`.
> 
> Since the `Other` (abbr: `XX`) value in all these properties are notions
of non-existance of breaking property, we want to switch to
`PartialCharProperty` domain type and use `Option<enum>`. This is left
as a separate step because it needs changes to the macro.

This assertion just isn't true! Nothing about `EnumeratedCharProperty` means it can't use `PartialCharProperty`. In fact, in the example/doctest for `char_property!(enum)`, we have the following:

https://github.com/behnam/rust-unic/blob/d51cf4f71b448d25a4f97c18df65128c4900be0f/unic/char/property/src/macros.rs#L43-L48

I think it works with `PartialCharProperty`. So here's a patch with it working with PartialCharProperty. Test failure is rustc:1.17.0 only, only where the target branch fails.

### DIFF
--- a/unic/ucd/Cargo.toml
+++ b/unic/ucd/Cargo.toml
@@ -20,10 +20,11 @@ unic-ucd-age = { path = "age/", version = "0.6.0" }
 unic-ucd-bidi = { path = "bidi/", version = "0.6.0" }
 unic-ucd-case = { path = "case/", version = "0.6.0" }
 unic-ucd-category = { path = "category/", version = "0.6.0" }
-unic-ucd-ident = { path = "ident/", version = "0.6.0" }
 unic-ucd-core = { path = "core/", version = "0.6.0" }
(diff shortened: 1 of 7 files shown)


CAD97:
This is a PR onto PR #166. Please note the header where it says

> [**CAD97**](https://github.com/CAD97) wants to merge 1 commit into `behnam:ucd-seg` from `CAD97:ucd-seg`

behnam:
Uh, looks like I mixed the responsibilities of the macro for Boolean vs Enumerated. You're right, for Enumerated types, we don't have the traits implemented and it's all in my code! I need to fix in the branch then and update the commit desc. Thanks, @CAD97 !

PR #174, opened 2017-10-06
Add unic-segment component

behnam:
Make all `unic-ucd-segment` properties Total, as turns out using these properties is easier as a total property, instead of partial, as we implemented initially. The reason for it is the fact that the default value (called `Other` in this context) is usually used in the same level as other values.

The algorithm implementation is based on [`unicode-segmentation`](https://github.com/unicode-rs/unicode-segmentation), with changes to use `unic-ucd-segment`.

The API is kept minimal at the moment, with only segment boundry objects and their indices objects. As we expand our string-level APIs, we can add on what's needed.

The tests for the algorithm are imported. Also new test is created for `unic-ucd-segment`, using the same source.

Closes <https://github.com/behnam/rust-unic/issues/135>


behnam:
bors: r+

PR #176, opened 2017-10-13
[ucd/normal/hangul] Use from_u32_unchecked

behnam:
Replace `mem::transmute()` with `char::from_u32_unchecked()`, as its
more clear on the semantics.

Also fix a broken integration test which is only run in special
conditions.

[travis] Add test step for unic-ucd-normal


behnam:
bors: r+

PR #179, opened 2017-11-06
[ucd] [char] [utils] Make libraries no_std

behnam:
Most of the transition was a matter of changing the namespace from `std` to `core`.

* Unit tests for `Display` implementations are moved into integration tests, allowing use of `format!()` in test code.

* [`char-range`] needed special attention, because of having an API using `std::collections::Bound` as input. Although the datatype is safe for core, it's part of `liballoc` APIs and therefore not available in `libcore`. Therefore, a new `std` feature is added to the package, which enables the part of the API.

* [`char-range`] Feature implementations are refactored into separate files to simplify configuration conditioning.

* [`char-property`] To be able to be used in `no_std` libraries, had to bring in some code for ASCII-case-insensitive char and str matching.

* [`char-property`] The `char_property!()` macro now depends on the caller package to include `extern crate core;` explicitly iff it's a normal (use-std) package.


behnam:
bors: r+

PR #183, opened 2017-11-09
[char/property] Merge in CharDataTable

behnam:
`CharDataTable` is a part of the character property definition and is
only used for character properties. So, it makes sense to keep it in
`unic-char-property` and drop the generic `unic-utils` component.


behnam:
bors: r+

behnam:
bors: r+

behnam:
bors: r+

PR #193, opened 2018-01-11
[char/property] Drop delimiter token needed for pre-1.20 rustc

behnam:
* Suppress AsciiExt unused_imports warning.

* [char/property] Drop delimiter token needed for pre-1.20 rustc.


behnam:
bors: r+

Issue #171, opened 2017-10-02
[ucd-name] Complete implementation for Unicode Name Property


behnam:
We have added a basic partial implementation for character names. We need a few more sources (data-based and algorithmic) to complete the implementation.

From <http://www.unicode.org/reports/tr44/#About_Property_Table>

> Jamo.txt must also be used, and the Name property for CJK unified ideographs, Tangut ideographs, and Nushu ideographs is derived by rule.

Source data:
* http://www.unicode.org/Public/UCD/latest/ucd/Jamo.txt

From the spec (http://www.unicode.org/versions/Unicode10.0.0/ch04.pdf) (Section *4.8 Name*), NR3 and NR4 are already implemented, and we still need implement these rules:
* NR1: Hangul Jamos
* NR2: Ideographics

From *Table 4-8. Name Derivation Rule Prefix Strings*:
```
Range       Rule Prefix String
AC00..D7A3   NR1 “hangul syllable”
3400..4DB5   NR2 “cjk unified ideograph-”
4E00..9FEA   NR2 “cjk unified ideograph-”
20000..2A6D6 NR2 “cjk unified ideograph-”
2A700..2B734 NR2 “cjk unified ideograph-”
2B740..2B81D NR2 “cjk unified ideograph-”
2B820..2CEA1 NR2 “cjk unified ideograph-”
2CEB0..2EBE0 NR2 “cjk unified ideograph-”
17000..187EC NR2 “tangut ideograph-”
1B170..1B2FB NR2 “nushu character-”
F900..FA6D*  NR2 “cjk compatibility ideograph-”
FA70..FAD9   NR2 “cjk compatibility ideograph-”
2F800..2FA1D NR2 “cjk compatibility ideograph-”
```

NOTE: *Code Point Labels*, as defined later in that chapter shall be implemented with its own API. Let's not include those in this issue, and we get to them after we're done here.

eyeplum:
It's nice to find this task here as I'm needing this feature right now :)
I have basic knowledge about Rust, and would like to give this a try.

For now I'm thinking about expanding the code point ranges and add `(code_point, derived_name)` to the [name table](https://github.com/behnam/rust-unic/blob/master/unic/ucd/name/tables/name_map.rsv). Will this be a proper way to do it?

Thanks!

CAD97:
[...]

PR #196, opened 2018-01-31
Implement Unicode Name - NR2

eyeplum:
This is a work in progress for implementing Unicode Name - NR2 (#171).

- [x] NR2: CJK Unified Ideograph / Tangut Ideograph / Nushu Character / CJK Compatibility Ideograph
- [x] Add tests

Feedback is welcome!

### DIFF
--- a/unic/ucd/name/src/name.rs
+++ b/unic/ucd/name/src/name.rs
@@ -9,6 +9,7 @@
 // except according to those terms.
 
 use core::fmt;
+use core::cmp::Ordering;
 
 pub static PREFIX_HANGUL_SYLLABLE: &'static str = "HANGUL SYLLABLE ";
 pub static PREFIX_CJK_UNIFIED_IDEOGRAPH: &'static str = "CJK UNIFIED IDEOGRAPH-";
@@ -16,7 +17,7 @@ pub static PREFIX_TANGUT_IDEOGRAPH: &'static str = "TANGUT IDEOGRAPH-";
 pub static PREFIX_NUSHU_CHARACTER: &'static str = "NUSHU CHARACTER-";
 pub static PREFIX_CJK_COMPATIBILITY_IDEOGRAPH: &'static str = "CJK COMPATIBILITY IDEOGRAPH-";
 
-#[derive(Copy, Clone, Debug, Ord, PartialOrd, Eq, PartialEq, Hash)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 pub enum Name {
     Pieces(&'static [&'static str]),
     HangulSyllable(char),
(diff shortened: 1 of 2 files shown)


eyeplum:
Thanks for the review!
I will update according to the comments and add some tests, also will put off `NR1` for now.
I already have some rough ideas about how `NR1` could be implemented, so hopefully a new PR about that will come soon :)
Thanks again.

unic/ucd/name/src/name.rs
```diff
@@ -10,37 +10,138 @@
...
+                let (first, rest) = pieces.split_first().unwrap();
+                f.write_str(first)?;
+                for piece in rest {
+                    f.write_str(" ")?;
+                    f.write_str(piece)?;
+                }
+                Ok(())
+            },
+            &Name::HangulSyllable(ch) => {
+                // FIXME: unimplemented
```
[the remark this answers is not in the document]
eyeplum: Added a temporary implementation with `HANGUL SYLLABLE {HEX_VALUE}`.

unic/ucd/name/src/name.rs
```diff
@@ -10,37 +10,138 @@
...
+            },
+            &Name::CJKCompatibilityIdeograph(ch) => {
+                let mut len = PREFIX_CJK_COMPATIBILITY_IDEOGRAPH.len();
+                len += self.num_of_digits(ch as u32);
+                len
+            },
+        }
+    }
+
+    fn num_of_digits(&self, code_point: u32) -> usize {
```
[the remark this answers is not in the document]
eyeplum: Fixed.

unic/ucd/name/src/name.rs
```diff
@@ -10,37 +10,138 @@
...
+    NushuCharacter(char),
+    CJKCompatibilityIdeograph(char),
 }
 
 #[cfg_attr(feature = "clippy", allow(len_without_is_empty))]
 impl Name {
     pub fn of(ch: char) -> Option<Name> {
-        data::NAMES.find(ch).map(|pieces| Name { pieces })
+        let code_point = ch as u32;
+        match code_point {
```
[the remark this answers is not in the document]
eyeplum: Fixed.

unic/ucd/name/src/name.rs
```diff
@@ -10,37 +10,138 @@
 
 use core::fmt;
 
+pub static PREFIX_HANGUL_SYLLABLE: &'static str = "HANGUL SYLLABLE ";
+pub static PREFIX_CJK_UNIFIED_IDEOGRAPH: &'static str = "CJK UNIFIED IDEOGRAPH-";
+pub static PREFIX_TANGUT_IDEOGRAPH: &'static str = "TANGUT IDEOGRAPH-";
+pub static PREFIX_NUSHU_CHARACTER: &'static str = "NUSHU CHARACTER-";
+pub static PREFIX_CJK_COMPATIBILITY_IDEOGRAPH: &'static str = "CJK COMPATIBILITY IDEOGRAPH-";
+
 #[derive(Copy, Clone, Debug, Ord, PartialOrd, Eq, PartialEq, Hash)]
```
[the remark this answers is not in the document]
eyeplum: Ended up with a big 2-level matching, happy to change it if there is a more concise way.

unic/ucd/name/tests/basic_tests.rs
```diff
@@ -13,12 +13,37 @@ extern crate unic_ucd_name;
 use unic_ucd_name::Name;
 
 #[test]
-fn basic_tests() {
+fn test_name_str() {
     assert_eq!(
         Name::of('A').expect("No name for A").to_string(),
```
behnam: nit: In tests, it's better to not use `expect()`, unless it's something that may fail. I think using `unwrap()` in these cases make the test more readable by being shorter, and more clean about expectation.

unic/ucd/name/tests/basic_tests.rs
```diff
@@ -13,12 +13,37 @@ extern crate unic_ucd_name;
...
+        Name::of('理').expect("No name for 理").to_string(),
+        "CJK COMPATIBILITY IDEOGRAPH-F9E4"
+    );
     assert_eq!(Name::of('\u{10FFFF}'), None);
+}
 
+#[test]
+fn test_name_cmp() {
     assert!(Name::of('A') < Name::of('B'));
+    assert!(Name::of('言') < Name::of('A'));
```
behnam: Would you please expand the test here to compare `Name::of()` AND their `to_string()` for any pair of sample?

unic/ucd/name/tests/basic_tests.rs
eyeplum: Sounds good, fixed :)

unic/ucd/name/tests/basic_tests.rs
```diff
@@ -13,12 +13,37 @@ extern crate unic_ucd_name;
...
 
 #[test]
-fn basic_tests() {
+fn test_name_str() {
     assert_eq!(
-        Name::of('A').expect("No name for A").to_string(),
+        Name::of('A').unwrap().to_string(),
         "LATIN CAPITAL LETTER A"
     );
+    assert_eq!(
```
behnam: Uh, another nit: It's good practice to group the asserts together based on their rule in the spec, and have a comment for each group with the rule name. Like this:
```
// == NR1: Hangul Jamos ==

assert_eq!(
    ...
)
assert_eq!(
    ...
)

// == NR2: Ideographics ==

assert_eq!(
    ...
)
assert_eq!(
    ...
)

// == NR3: ... ==

assert_eq!(
    ...
)
assert_eq!(
    ...
)
```

unic/ucd/name/tests/basic_tests.rs
eyeplum: Expanded the test with more comparison between samples and their `to_string()` comparison.

unic/ucd/name/tests/basic_tests.rs
eyeplum: Fixed.

eyeplum:
Thanks for the review @behnam!
I have addressed all the comments and rebased to the latest `master`.
I have also run `cargo fmt` and committed the changes.
Although I'm not able to get `cargo clippy` to work - tried to install with `cargo +nightly install clippy`, got compile errors :(

unic/ucd/name/tests/basic_tests.rs
```diff
@@ -13,12 +13,170 @@ extern crate unic_ucd_name;
...
     assert_eq!(Name::of('\u{10FFFF}'), None);
+}
+
+#[test]
+fn test_name_cmp() {
+    // == NR1 ==
+
+    // FIXME: use decomposed jamo short names instead
+    assert!(Name::of('곲') > Name::of('가'));
+    assert!(Name::of('곲').unwrap().to_string() > Name::of('가').unwrap().to_string());
```
behnam: Feel like writing a little macro here for the test logic, and only call the macro on test values?

Here are two examples of doing that in our tests:

https://github.com/behnam/rust-unic/blob/c2ed9f869964192f96ac137fc87847aa7cbb1ca5/unic/segment/tests/words_conformance_tests.rs#L28-L39

https://github.com/behnam/rust-unic/blob/c2ed9f869964192f96ac137fc87847aa7cbb1ca5/unic/ucd/common/tests/consistency_tests.rs#L20-L39

eyeplum: Fixed :)

eyeplum:
Thanks @behnam for the advice. 
A macro indeed makes the tests look better and easier to change.
I have also squashed a few commits.

CAD97:
(delegate means you have permission to `bors: r+` it now, when you're ready)

CAD97:
bors: delegate+

eyeplum:
@CAD97 sure, I will explore and see.
One other thing I'm thinking about is: since `len()` `fmt()` and `cmp()` are all based on the string value of `self`, is it also possible to simply create a `string_value()` method and call it from those methods? Not sure about the memory footprint though.
I notice the initial implementation was going this way but 1a50398 changed it.

CAD97:
The trick is to avoid creating the `String` if unnecessary. Because `Display` gives `ToString` for free, the use-site can create a `String` that way if they want it.

Also note that this crate is `#[no_std]`. We want to keep these libraries available to no-std environments as much as possible.

eyeplum:
Ah, make sense. Thanks!

unic/ucd/name/src/name.rs
```diff
@@ -9,38 +9,234 @@
...
+            }
+            &Name::CJKCompatibilityIdeograph(ch) => {
+                let mut len = PREFIX_CJK_COMPATIBILITY_IDEOGRAPH.len();
+                len += self.num_of_hex_digits(ch as u32);
+                len
+            }
+        }
+    }
+
+    fn num_of_hex_digits(&self, code_point: u32) -> usize {
```
behnam: You can write this as:

```rust
fn number_of_hex_digits(ch: char) -> usize {
    (32 - u32::leading_zeros(ch as u32) as usize + 3) / 4
}
```

Please note that it should be a static method, since `&self` is not needed in the impl. (`clippy` would have caught this, I think.)

CAD97: (I knew there existed a clever solution but I couldn't think of it)

behnam: You'll get to master leading-zero/trailing-zero kinds of questions after school, in job interviews! ;)

behnam: Oh, btw, @eyeplum, either make this a macro, or `#[inline=always]` it. There's no reason for it not being inline.

behnam: Here's the `libcore` implementation of (still unstable) `char::escape_unicode`, which uses the same trick, but a bit more readable:

https://github.com/rust-lang/rust/blob/ddaab61101e48f0a58ce176b5a6aff11ae36c986/src/libcore/char.rs#L416-L432

For more fun like this, see this list: https://github.com/rust-lang/rust/search?utf8=%E2%9C%93&q=leading_zeros&type=

And, deep down, this is the impl: https://github.com/rust-lang/rust/blob/a538fe7ce715c7bd27e2e05329c3d857b9ad92af/src/libcore/num/mod.rs#L1357

Which calls LLVM's `ctlz`: https://llvm.org/docs/LangRef.html#llvm-ctlz-intrinsic

eyeplum: Fixed, thank you!

eyeplum:
@CAD97 Tried with `enum Name { NR1, NR2, NR3 }`, it does simplify things a lot! So let do it this way!
Thanks for the advice!

eyeplum:
bors: r+

unic/ucd/name/src/name.rs
```diff
@@ -9,38 +9,138 @@
...
-pub struct Name {
-    pieces: &'static [&'static str],
+pub static PREFIX_HANGUL_SYLLABLE: &'static str = "HANGUL SYLLABLE ";
+pub static PREFIX_CJK_UNIFIED_IDEOGRAPH: &'static str = "CJK UNIFIED IDEOGRAPH-";
+pub static PREFIX_TANGUT_IDEOGRAPH: &'static str = "TANGUT IDEOGRAPH-";
+pub static PREFIX_NUSHU_CHARACTER: &'static str = "NUSHU CHARACTER-";
+pub static PREFIX_CJK_COMPATIBILITY_IDEOGRAPH: &'static str = "CJK COMPATIBILITY IDEOGRAPH-";
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub enum Name {
```
behnam: This looks more clean now! Thanks, @eyeplum, for the update.

If you like working more on this component, would be great to add some documentation here with links to the specification of NR1, etc.

Here are two examples:

https://github.com/behnam/rust-unic/blob/master/unic/ucd/age/src/age.rs#L16-L39

https://github.com/behnam/rust-unic/blob/master/unic/ucd/bidi/src/bidi_class.rs#L17-L23

Also, you can add
```rust
#![deny(missing_docs)]
```
to `src/lib.rs` of this component, to require documentation on all public items on compile time.

eyeplum: Sure! I will be working on `NR1` soon and will add the documentations.
Thank you so much for your guidance!

PR #200, opened 2018-02-05
Add gh-pages doc upload to nightly build

CAD97:
@behnam you'll need to add a `$GH_TOKEN` to Travis for this to work.

> In order to generate a GitHub token, navigate to [your tokens settings page](https://github.com/settings/tokens). While you're there, review your currently issued tokens, and revoke those that you're no longer using. Generate a new token, give it a descriptive name (this can be edited later) (I suggest `repo-name-doc-upload`) and the `public_repo` scope (this can be edited later). Make sure to copy the token once it's generated, as you will never be able to see it again (though you can regenerate it)! This token gives anyone with the token access to act as you when read/writing to any public repository you have the permission to read/write to, so keep it safe. Add it to the GH_TOKEN environment variable on Travis and _make sure that "Display value in build log" is set to off_.
>
> - [CAD97 on Great Rust CI](https://dev.to/cad97/great-rust-ci-1fk6)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,10 +46,9 @@ script:
 
 # TODO: - rustdoc --test README.md -L target/debug -L target/debug/deps
 
-# TODO: Enable auto update for gh-pages
-#after_success: |
-#  cargo doc && \
-#  echo '<meta http-equiv="refresh" content="0;url=YOURLIBNAME/index.html">' > target/doc/index.html && \
-#  sudo pip install ghp-import && \
-#  ghp-import -n target/doc && \
-#  git push -qf https://${GITHUB_TOKEN}@github.com/${TRAVIS_REPO_SLUG}.git gh-pages
+after_success:
+  # upload gh-pages documentation
+  - test "$TRAVIS_RUST_VERSION" != "nightly" ||
+    (cargo doc --verbose --all --all-features &&
+     cargo install cargo-travis &&
+     cargo doc-upload --branch master)


behnam:
Thanks, @CAD97, for the diff!

This TODO item was left over from an old project, IIRC. With <https://docs.rs/unic/> hosting the docs for the latest release, do you think it still worth setting up a CI-based doc publication for master?

Also, last time I checked (in the summer), the `cargo-travis` package was not maintained actively anymore and didn't work in some cases that I needed. Do you think it's going to be reasonably stable to enable it here?

behnam:
And something else related here is that we have other kinds of docs besides API docs, currently put under <https://github.com/behnam/rust-unic/tree/master/docs>.

My vision is to have more of these kinds of docs, as many Unicode & i18n issues are not just API docs, but architecture and which APIs need to be used for what application.

So, I was hoping that, whenever we set a homepage-kind-of-thing up for UNIC, it will cover both of these docs.

If we want to do this, then we need to compose the root `README.md`, with the `docs/` dir and `target/doc/` into on structure, and put that into `gh-pages`.

What do you think?

CAD97:
Ah right. I forgot about those docs.

Note that [travis-cargo](https://github.com/huonw/travis-cargo) (the python script) is different from [cargo-travis](https://github.com/roblabla/cargo-travis), to which I merged a couple PRs to enable doc-upload and a couple other things.

If we're already serving docs from the `/docs` folder, then we're good as is (feel free to close this). `master` docs are useful for some crates, whereas docs.rs only serves the published versions. It also would be slightly more useful potentially given that all of the crates would be available in one flat hierarchy and selectable on the left menu.

If we want have the API docs built and the guide-like docs from `/docs` built to gh-pages, I can spin something together next weekend probably. The files that doc-upload touches are specified clearly in order to enable something like this.

CAD97:
Closing for now. I'll investigate building guide- and reference- docs in the future, especially now that I've broken out `cargo-ghp-upload` which I know should work properly now.

PR #203, opened 2018-02-07
Release v0.7.0

behnam:
# UNIC Applications

UNIC Applications are binary creates hosting in the same repository as `unic` super-crate, under the `apps/` directory. These creates are not internal parts of the `unic` library, but tools designed and developed for the general audience, also serving as a test bed for the UNIC API. We are starting with CLI applications, and possibly expanding it to GUI and WEB applications, as well.

- [`unic-cli`] The new UNIC CLI application provides command-line tools for working with Unicode characters and strings. In this release, first versions of `unic-echo` and `unic-inspector` commands are implemented.

# New Components

## Character Property

- [`unic-ucd-common `] Common character properties (*alphabetic*, *alphanumeric*, *control*, *numeric*, and *white_space*).
- [`unic-ucd-ident`] Unicode Identifier character properties.
- [`unic-ucd-segment`] Unicode Segmentation character properties.
- [`unic-emoji-char`] Unicode Emoji character properties.

## String Algorithm

- [`unic-segment`] Implementation of Unicode Text Segmentation algorithms (*Grapheme Cluster* and *Word* boundaries).

# Other Updates

This release was delayed for a couple of cycles, because of the problems with running tests in a workspace with a mix of std and no-std creates. The issue is resolved as of `1.22.0`.

- Enable `no_std` for many of the existing components.
- Bumped minimum Rust to `1.22.0`.
- Lots of small fixes for data types and internal structure updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ In `Cargo.toml`:
 
 ```toml
 [dependencies]
-unic = "0.6.0"  # This has Unicode 10.0.0 data and algorithms
+unic = "0.7.0"  # This has Unicode 10.0.0 data and algorithms
 ```
 
 And in `main.rs`:
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ this crate.
 
 ### Major Components
 
--   [`unic-char`](char): Unicode Character Tools.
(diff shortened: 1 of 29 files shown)


behnam:
bors: r+

PR #204, opened 2018-02-09
Add lint attributes to components

behnam:
- Add lint attribute checker for components, to make sure some basic lint rules are enabled everywhere in the library.

- Add missing lint attributes to components. Add FIXME where more work is needed.

- Enable building benches targets, without running them, so the code doesn't rust over time.


behnam:
bors: r+

PR #208, opened 2018-02-22
Implement Unicode Name - NR1

eyeplum:
This pull request includes the implementation for Unicode Name - NR1 (#171).

- [x] NR1: Hangul Syllables
- [x] Add documentations for `unic-ucd-name`
- [x] Generate `jamo.rsv` from `Jamo.txt`


CAD97:
Looks decent, I'll do a proper review over the weekend when I've got a bit of time.

[This regex](https://regex101.com/r/pIx6E8/1) grabs the information that you need:

```regex
(?xm)^ # every line
([[:xdigit:]]{4,6}) # codepoint
;[[:blank:]]*
([[:alpha:]]*) # jamos
```

And you're going to want to write it out as a [DirectCharTable](https://github.com/behnam/rust-unic/blob/master/gen/src/writer/utils/tables/direct.rs)

unic/ucd/name/src/name.rs
```diff
@@ -11,16 +11,41 @@
 use core::cmp::Ordering;
 use core::fmt;
 
+extern crate unic_ucd_normal;
```
behnam: I believe `extern crate` is expected to live in `lib.rs`.

unic/ucd/name/src/name.rs
```diff
@@ -74,21 +101,49 @@ impl Name {
...
     }
+
+    fn jamo_short_name(ch: char) -> &'static str {
+        data::JAMO_SHORT_NAMES
+            .find(ch)
+            .expect("Unexpected jamo character")
+    }
+
+    fn collect_jamo_short_names(ch: char) -> [&'static str; 3] {
+        let mut jamos = [""; 3];
```
behnam: Curious if using an *empty string* is the best way of doing this here. Any thoughts?

unic/ucd/name/src/name.rs
```diff
@@ -148,4 +207,5 @@ mod data {
     use unic_char_property::tables::CharDataTable;
     include!("../tables/name_values.rsd");
     pub const NAMES: CharDataTable<&[&str]> = include!("../tables/name_map.rsv");
+    pub const JAMO_SHORT_NAMES: CharDataTable<&str> = include!("../tables/jamo.rsv");
```
behnam: Considering the fact that these values are not that useful directly, how about we just make both of these private?

unic/ucd/name/tests/basic_tests.rs
```diff
@@ -62,7 +59,6 @@ fn test_name_cmp() {
 
     // == NR1 ==
 
-    // FIXME: use decomposed jamo short names instead
```
behnam: Should we add more tests here for sort ordering of JAMO vs JAMO and JAMO vs others?

unic/ucd/normal/src/lib.rs
```diff
@@ -54,6 +54,7 @@ mod decomposition_type;
 pub use decomposition_type::DecompositionType;
 
 mod hangul;
+pub use hangul::decompose as decompose_hangul_syllable;
```
behnam: Would you please add a `TODO(GH-209): Drop after unic-ucd-hangul is available` here?

And please feel free to grab it yourself, or post it to This Week in Rust for new contributors.

unic/ucd/name/src/name.rs
eyeplum: Fixed!

unic/ucd/name/src/name.rs
eyeplum: Changed the array to type `Option<&'static str>` with 3 elements so we can use `None` as the default values.
Feel like this makes more sense.

unic/ucd/name/tests/basic_tests.rs
eyeplum: Sure, will add a few more tests!

unic/ucd/name/src/name.rs
eyeplum: I tried to remove the `pub`, but looks like it will make these constants non-accessible from `Name` :(

behnam: Oh, these are only `pub` in the private `mod data`. So, yeah, need to be `pub` and is not exposed to the outside of the crate, meaning it's fine. Thanks for looking!

unic/ucd/name/src/name.rs
```diff
@@ -47,19 +70,21 @@ impl Name {
...
     pub fn len(&self) -> usize {
-        match self {
-            &Name::NR1(ch) => {
+        match *self {
+            Name::NR1(ch) => {
                 let mut len = PREFIX_HANGUL_SYLLABLE.len();
-                // FIXME: use decomposed jamo short names instead
-                len += Name::number_of_hex_digits(ch);
+                {
+                    let mut count_jamos = |_| len += 1;
```
CAD97: Some of the jamo short names in the table are more than one character long -- is this correct?

(I really have no idea how hangul is handled in Unicode...)

eyeplum: Nice catch!
Yeah, some of them are more than one and some of them are less than one (empty).
I will change this to use the length of the jamo.

unic/ucd/name/src/name.rs
```diff
@@ -78,21 +106,49 @@ impl Name {
...
+            .find(ch)
+            .expect("Unexpected jamo character")
+    }
+
+    fn collect_jamo_short_names(ch: char) -> [Option<&'static str>; 3] {
+        let mut jamos = [None; 3];
+        {
+            let mut index = 0;
+            let mut collect_jamos = |jamo| {
+                jamos[index] = Some(Name::jamo_short_name(jamo));
```
CAD97: Can we add a (`debug_`?)`assert!` here for the bound on `index`? If max number of short names turns out to be higher it'd be better to have an assertion fail and say what overflowed rather than just an out-of-index, even if it's pointing here.

eyeplum: Added a `debug_assert!`, thanks!

eyeplum:
bors: r+

PR #214, opened 2018-03-05
Implement new unic-char-basics component

behnam:
The new `unic-char-basics` component  provides basic Unicode Character tools and properties that are core to string processing that are [stabilized](http://www.unicode.org/policies/stability_policy.html) by The Unicode Standard and will never change in the future.


behnam:
bors: r+

Issue #209, opened 2018-02-22
[unic-ucd-hangul] Hangul Syllable Composition/Decomposition


behnam:
Let's move the Hangul Syllable Composition/Decomposition logic out of `unic-ucd-normal` and put it in its own standalone component, to be used out of normalization context.

Here's the impl: https://github.com/behnam/rust-unic/blob/master/unic/ucd/normal/src/hangul.rs

eyeplum:
I will pick up this one :)

eyeplum:
@behnam I think this one can be closed :)

PR #217, opened 2018-03-10
Move Hangul Syllable composition and decomposition to [unic-ucd-hangul]

eyeplum:
Implement #209


unic/ucd/hangul/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "unic-ucd-hangul"
+version = "0.7.0"
+authors = ["The UNIC Project Developers"]
+repository = "https://github.com/behnam/rust-unic/"
+license = "MIT/Apache-2.0"
+description = "UNIC — Unicode Character Database — Hangul"
```
behnam: Could you please elaborate on what the create is here?

Maybe "Hangul Syllable Composition & Decomposition", to follow the name of the algorithms in TUS (The Unicode Standard)?

unic/ucd/hangul/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+// except according to those terms.
+
+#![no_std]
+#![forbid(future_incompatible, missing_debug_implementations, unconditional_recursion, unsafe_code)]
+#![deny(bad_style, unsafe_code, unused)]
+
+extern crate unic_ucd_version;
+use unic_ucd_version::UnicodeVersion;
+
+pub mod hangul;
```
behnam: Since `hangul` algorithms are all this component providing, let's make the module here private and expose the API via `pub use`.

Also, what do you think about renaming the public API functions to `compose_syllable()` and `decompose_syllable()`.

IMHO, it would reflect the algorithm better, and is more readable.

unic/ucd/hangul/tests/basic_tests.rs
```diff
@@ -0,0 +1,53 @@
...
+        let mut collect_decomposed = |chr| {
+            decomposed_lvt.push(chr);
+        };
+        hangul::decompose('퓛', &mut collect_decomposed);
+    }
+    assert_eq!(decomposed_lvt, ['ᄑ', 'ᅱ', 'ᆶ']);
+}
+
+#[test]
+fn test_compose() {
```
behnam: I missed this test file before. Looks good. Maybe we can add more function to it that decomposes and the composes, and use that in the docblock tests?

(Or you can leave the test additions fro another PR. Up to you.)

unic/ucd/hangul/Cargo.toml
eyeplum: Fixed :)

unic/ucd/hangul/src/lib.rs
eyeplum: Agreed, refined the APIs.

unic/ucd/hangul/tests/basic_tests.rs
eyeplum: Added a new test case `test_decompose_then_compose` for this.
Also used the new test case in the docblock tests.

eyeplum:
Thanks for the review @behnam!
I have updated the PR according to the suggestions.
Also added a simple `README.md` file.

behnam:
bors: r+

PR #221, opened 2018-04-15
Rayon ParallelIterator support for CharRange

CAD97:
Also, could you see if https://github.com/rust-lang/cargo/issues/5362 happens for you?


behnam:
Awesome! Thanks, @CAD97!

No, the bug doesn't repro on my system with nightly rust. So will go ahead with this.

bors: r+

Issue #224, opened 2018-06-07
Implement unic-ucd-unihan


eyeplum:
Provides access to:
- `Unihan_DictionaryIndices`
- `Unihan_DictionaryLikeData`
- `Unihan_IRGSources`
- `Unihan_NumericValues`
- `Unihan_OtherMappings`
- `Unihan_RadicalStrokeCounts`
- `Unihan_Readings`
- `Unihan_Variants`

References:
- https://www.unicode.org/reports/tr38/tr38-25.html

eyeplum:
Please refer to #225 for a possible way to implement this.

PR #225, opened 2018-06-07
Sample implementation for unic-ucd-unihan

eyeplum:
This is a sample implementation of `ucd-unihan` #224 .
This implementation works in some way (used by [this app](https://itunes.apple.com/us/app/cicero-unicode-tool/id1372549464)), but may need major redesigns to fit into `rust-unic`.

One other thing to consider is - as Unihan is a CJK centric module in the Unicode standard - maybe we could make this crate an optional subcrate of the `rust-unic` super crate and user needs to opt-in explicitly to use it.


CAD97:
Failures are unrelated; I'll fix those as soon as I get a chance this weekend.

At a glance-over, this looks good; I'll do a more detailed pass this weekend.

behnam:
Thanks for building this, @eyeplum!

I'm in the process of moving source data files out of the repository, to be imported as submodules. (The download+unzip scripts will be in Python, I guess.)

Having the data externally, we won't have to deal with downloading/unzipping ourselves, which would be much better for this repo. Also, allowing easier addition of other sources and models.

If you like, we can rebase and try to land this work, and drop the data retrieving parts later. Or, we can just wait for the external sources and drop the data source work from this PR. What do you think?

eyeplum:
Sure, wait for the external sources sounds like a better option 👍

LuoZijun:
Ping ?

behnam:
In #247, I have added the complete Unicode UCD data package, under the new address: `/external/unicode/ucd/data/`, which also contains a `Unihan/` directory with all of its txt files. (See https://github.com/open-i18n/data-unicode-ucd/tree/master/data/Unihan)

So, there's no `data` package anymore, and you just need to write the `gen` rules for the tables, and implement the algorithm in a new component.

Also, as a reminder, since the new source data files are imported as submodules, you need to do `git submodule update --init` to get the files, before running `/etc/generate-tables.sh`.

eyeplum:
Hey @behnam , thanks for the review!
I have updated the pull request to read Unihan data from `./external/unicode/ucd/data/Unihan/` and migrated related code changes to Rust 2018.

eyeplum:
Hey @behnam , I'm planning to merge this in this weekend. Although the functionalities are very limited at the moment, I imagine this might be somewhat useful for users that needs Unihan. I'm planning to map most of the Unihan tables later this year, hopefully I will have enough time to make it happen.

As of now, before I start mapping more Unihan contents, I'm thinking about tackling the Unicode 11.0 upgrade for rust-unic first. Mainly because Unicode 12.0 is coming and I think it would be nice for us to at least update to Unicode 11.0 so future updates will be more manageable. I may need some help planning the work as there seems to be a lot involved. I will probably create a new issue so we can have more detailed discussions there.

What do you think?

unic/Cargo.toml
```diff
@@ -27,7 +28,7 @@ unic-emoji = { path = "emoji/", version = "0.9.0" }
 unic-idna = { path = "idna/", version = "0.9.0" }
 unic-normal = { path = "normal/", version = "0.9.0" }
 unic-segment = { path = "segment/", version = "0.9.0" }
-unic-ucd = { path = "ucd/", version = "0.9.0" }
+unic-ucd = { path = "ucd/", version = "0.9.0", features = [] }
```
eyeplum: Not sure if this is the best way to do it, the indention is to make the `unihan` feature turned-off by default.
Currently the `unic-ucd` only have a `unihan` feature, so this is probably fine, if we start to add more features to `unic-ucd`, then we will need to explicit specify the features here if we want the feature to be opt-in by default.

behnam: I think it's better to have `default = [ <enabled-features> ]` on the `unic-ucd` side. If there's no other enabled-by-default feature there, I think we can rely on the default value for `default = []` then.

So, just create a `[features]` item for unihan and don't put `unihan` in the list of deps?

eyeplum: Right, `unic-ucd = { path = "ucd/", version = "0.9.0" }` should work, I've removed `features = []`.

I've verified with an example project that `cargo build` won't build `unic-ucd-unihan`, and `cargo build --feature unihan` will. Running `cargo build` in `./unic` or `./unic/ucd` works in the same way.

I thought it was not working because when I ran `cargo build` in the root of the project it built `unic-ucd-unihan` (any ideas why?).

CAD97: Since what we have is a "virtual workspace" (that is, the root `Cargo.toml` is has a `[workspace]` table and no `[package]` table) `cargo build` in the root directory defaults to `cargo build --all`.

eyeplum: Ah, had a read through https://doc.rust-lang.org/cargo/reference/manifest.html#virtual-manifest now it makes more sense, thanks!

eyeplum:
Hey @behnam and @CAD97 , are you happy if I merge this in?

mozillazg:
Is there any news?

asg0451:
Is this still being worked on?

eyeplum:
@asg0451, I'm not planning to merge this anytime soon, you could try it out in my fork if you are interested in Unihan https://github.com/eyeplum/rust-unic

Issue #29, opened 2017-07-13
[ucd/normal] Characters 가 through 힣 have the wrong Decomposition_Type


CAD97:
See #27 which adds a failing test.

```
11172 test cases failed! (1100892 passed) {

0: Fail { line_num: Some(234), char: '가', exp_dt: Some(Canonical), actual_dt: None }

...

11171: Fail { line_num: Some(234), char: '힣', exp_dt: Some(Canonical), actual_dt: None }

}
```

Relevant line from [DecompositionType.txt]:

```
AC00..D7A3    ; Canonical # Lo [11172] HANGUL SYLLABLE GA..HANGUL SYLLABLE HIH
```

[DecompositionType.txt]: http://www.unicode.org/Public/UCD/latest/ucd/extracted/DerivedDecompositionType.txt

behnam:
Great find! Yeah, Hangul characters have a special handling that's done algorithmically and not based on data files. We need to implement it.

behnam:
@calum, this is good starter task. What you need to do is to use the const values in <https://github.com/behnam/rust-unic/blob/master/unic/ucd/normal/src/hangul.rs> to add a check for `DecompositionType` implementation, to return the correct value for Hangul chars.

To test your work, you need to remove the `#[should_panic(...)]` line from the conformance test (<https://github.com/behnam/rust-unic/blob/master/unic/ucd/normal/tests/conformance_tests.rs#L34>), expecting all the tests to pass.

What do you think?

calum:
Sounds good! I'll get to work. Thanks for the tips.

behnam:
@calum, this is the main code-related issue blocking UNIC-0.5 release. Do you think you can submit a PR today or early tomorrow, or would you mind if I fix this myself and find you a new issue next week?

behnam:
Btw, sorry to not mention that before. I just set up the milestone tags for all major releases, so we can better communicate these matters.

calum:
Hi, sorry I won't be able to get it finished this weekend. I've had family visiting this week so I haven't had time unfortunately. 

I'll find another issue to work on next week.

behnam:
No worries, @calum. I'll fix this one and we find you new issues next week. Thanks anyway! :)

behnam:
Fixed by <https://github.com/behnam/rust-unic/issues/72>

PR #226, opened 2018-07-31
Partial update to unicode 11

Alexendoo:
I was attempting to update the repo to use unicode 11 data but ran into a few issues, for the time being I'm setting it aside since it was a bigger job than I was expecting, but I opened this in case it's helpful to save some time for anybody attempting to do the same.

Outstanding issues:

#### unic-segment - https://unicode.org/reports/tr29/#Modifications

This was the part that tripped me up the most, you can see my attempt to tackle it in https://github.com/Alexendoo/rust-unic/compare/unicode-11-partial...unicode-11. I think the forward word boundaries *might* be correct, but the others are not.

#### unic-ident - https://www.unicode.org/reports/tr31/#Modifications

There was a few changes to this spec, I didn't get round to seeing if any code changes are required but the tests do pass, so I don't know if anything is needed to be done

#### IDNA conformance

With the IDNA conformance test updated to use [IdnaTestV2.txt](https://www.unicode.org/Public/idna/11.0.0/IdnaTestV2.txt) I was able to add a test for `unic_idna::to_unicode`, however it doesn't return an `Err` when a status of `X4_2` is expected - in this PR the test will fail but it can be easily added to be ignored as `V2` and `C...` are if that's intentional behaviour

What this PR does accomplish:

- The data, with the exception of unic-ucd-segment and unic-segment are all updated to unicode 11
- Temporarily `#[ignore]`s the test for unicode version of unic-ucd-segment
- Adds a char_property for the new Extended_Pictographic emoji property
- Updates the IDNA conformance test to use the new format
- A small change to the unic-gen segmentation test generator, the format of the test file comments changed slightly
- The grapheme conformance tests were updated to be more like the word boundary conformance tests

---
This change is [Reviewable](https://reviewable.io/reviews/open-i18n/rust-unic/226)


gen/src/writer/ucd/segment_tests.rs
```diff
@@ -32,10 +32,20 @@ fn emit_grapheme_cluster_break_test_data(dir: &Path) {
...
 
+
         for (i, ch) in chars.iter().enumerate() {
-            let ref gcb = char_gcbs[i];
+            let gcb = match char_gcbs[i].as_str() {
+                // Corresponds to \p{Extended_Pictographic}, not an actual grapheme cluster property
+                "ExtPict" => continue,
+
+                "Extend_ExtCccZwj" => "Extend",
+                "ZWJ_ExtCccZwj" => "ZWJ",
```
eyeplum: 👍 for this. I had this issue when I was trying to get my fork working for Unicode 11.0, this is a clever fix!

Out of curiosity, there seems to be no documentation about Extend_ExtCccZwj and ZWJ_ExtCccZwj (it seems they only appear in grapheme cluster break test data). Any idea of the purpose of these two values?

I noticed @behnam was also asking about this in http://www.unicode.org/review/pri372/ .

Alexendoo: I'm not entirely sure where these names came from, in the case of `ZWJ` I noticed that all of them were renamed to `ZWJ_ExtCccZwj` so it seemed to make sense to just map the value back to `ZWJ`

In the case of `Extend` though there are both `Extend` and `Extend_ExtCccZwj` in the 11.0 test file,  the choice of which are used seems to depend on the codepoint and not the GB# rule. A bit of grepping + sort showed the different rules and character combos

```
[9.0] EMOJI MODIFIER FITZPATRICK TYPE-6 (Extend)
[0.2] COMBINING GRAPHEME JOINER (Extend)
[4.0] COMBINING GRAPHEME JOINER (Extend)
[9.0] COMBINING GRAPHEME JOINER (Extend)

[4.0] COMBINING DIAERESIS (Extend_ExtCccZwj)
[9.0] COMBINING DIAERESIS (Extend_ExtCccZwj)
[0.2] COMBINING GRAVE ACCENT (Extend_ExtCccZwj)
[4.0] COMBINING GRAVE ACCENT (Extend_ExtCccZwj)
[9.0] COMBINING GRAVE ACCENT (Extend_ExtCccZwj)
```

I don't know why they're distinguished myself, they're all still `Extend` in https://www.unicode.org/Public/11.0.0/ucd/auxiliary/GraphemeBreakProperty.txt

behnam:
Thanks, @Alexendoo, for submitting this!

We had build issues on the CI. Could you please rebase this so we can see if there are any problems?

Also, they were some non-straight-forward changes in UAX #29 in Unicode 11.0.0 release. (https://unicode.org/reports/tr29/#Modifications) How are we incorporating that to the implementation here?

Alexendoo:
@behnam this PR doesn't handle it, I spent some time on it but I didn't really get anywhere

The tests are expected to fail with expected `X4_2` in `unic_idna::to_unicode` since I wasn't sure if that was intentional or not (it came from added test coverage, not a change in implementation)

I don't know if you'd really want to merge this directly since it's not complete. It's more for visibility if somebody else wants to build the tr29 changes on top of it it will save them a bit of time

Issue #206, opened 2018-02-16
Implement unic-ucd-block


behnam:
Provides access to char property:
* `Block` (Enumerated)

Also let's have two iterators:
* All assigned/defined Unicode Blocks
* A character iterator that only walks on character ranges inside these blocks. This is very helpful for reducing the range of chars we test in integration tests, which is wasting a lot of CPU also walking on the ~75% of the code space that's not assigned.

References:
* http://www.unicode.org/reports/tr44/tr44-4.html#Script
* http://www.unicode.org/Public/UCD/latest/ucd/Blocks.txt

Consistency tests:
* Any character assigned should have a block.
* Some blocks contain unassigned characters.

Cross-component tests:
* IIRC, each Block can contain characters from one non-Common/Unknown Script. (GH-137)

Similar crates:
N/A

eyeplum:
While #217 is waiting for code review, I will try to implement this one.

behnam:
Thanks, @eyeplum! I already have a half-baked implementation on my laptop. If you haven't started yet, maybe wait a few days before you start?

eyeplum:
@behnam Sure!

behnam:
Thanks, @eyeplum, for wrapping this up! 😄

PR #227, opened 2018-08-01
Implement iterator for all assigned Unicode Blocks

eyeplum:
This pull request partially implements #206 by implementing a block iterator for all assigned Unicode Blocks.


eyeplum:
Hey @behnam I know you were working on #206 , but I needed this in my project so I also tried to implement this. Feel free to close this PR if it can be covered by your work, or take the useful parts.
Cheers!

eyeplum:
Thanks @behnam , I will rebase and update the PR!

unic/ucd/block/tests/basic_tests.rs
```diff
@@ -0,0 +1,30 @@
...
+fn test_block_of() {
+    let latin_1 = Block::of('\u{00a9}').unwrap(); // ©, COPYRIGHT SIGN
+    assert_eq!(latin_1.range.low, '\u{0080}');
+    assert_eq!(latin_1.range.high, '\u{00ff}');
+    assert_eq!(latin_1.name, "Latin-1 Supplement");
+}
+
+#[test]
+fn test_block_iter() {
+    assert_eq!(BlockIter::new().count(), 277);
```
eyeplum: Just found out that `char` can't represent surrogate code points. So there are 3 blocks missing when we generate the tables (Unicode 10.0 has 280 assigned blocks).

```
D800..DB7F; High Surrogates
DB80..DBFF; High Private Use Surrogates
DC00..DFFF; Low Surrogates
```

behnam: Right. And that's fine, since non of these codepoints are valid characters anymore. (They were only valid before introduction of UTF-16.)

unic/ucd/block/src/block.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+impl Block {
+    pub fn of(chr: char) -> Option<Block> {
+        Block::new(data::BLOCKS.find_with_range(chr))
+    }
+
+    fn new(raw_block: Option<(CharRange, &'static str)>) -> Option<Block> {
+        match raw_block {
+            None => None,
```
behnam: Hum... why do we need to be able to call `new()` with `None` and receive `None`. It's kind of counter-intuitive to Rust's `new()` pattern.

eyeplum: Good point, I think I was trying optimize for the two use cases in this file which both provide an `Option<(CharRange, &'static str)>` and expect an `Option<Block>`.
Should be fixed now.
I have also added some comments about the missing surrogate code point blocks.

behnam:
Awesome! Looks great!

bors: r+

behnam:
Oh, we need to fix `master` on `beta`/`nightly` first. A bunch of non-top-level `use` statements need a `create::` prefix. (See https://travis-ci.org/open-i18n/rust-unic/jobs/439036265) Would be great if you could help with it. If not, I'll try to get to it later this week.

eyeplum:
Hey @behnam, I tried to fix those warnings for `beta` and `nightly` in #233, but looks like the preferred way is not supported by `stable` and `1.26.0`, so I'm afraid we can't fix those warnings at the moment.
The build failure in this pull request turns out to be some outdated lint annotations in `ucd/block`'s `lib.rs` (I copied it from some other module at some point). I have aligned `ucd/block`'s lint annotations with other modules, now this pull request should build 🤞.

behnam:
Copying from the other PR (https://github.com/open-i18n/rust-unic/pull/233):

> Oh, I thought `master` is broken. Apparently it's not, but the builds here were failing because of the new code was based on the code before we allow warnings again.
> 
> So, let's just `crate::` prefixes and all should be fine now. Right?

PR #229, opened 2018-08-19
Fix compile errors and update project metadata

behnam:
Bumping minimum-rust-version to 1.24.0, because of compile failures on some dependencies.

Updating all docs/badges links to the new home: <https://github.com/open-i18n/rust-unic>.


behnam:
bors: r+

behnam:
bors: r+

PR #233, opened 2018-10-10
Fix path warnings in `use` declarations for beta and nightly

eyeplum:
This pull request fixes build warnings related to module paths in `use` declarations for beta and nightly toolchains.

### DIFF
--- a/unic/bidi/src/bidi_info.rs
+++ b/unic/bidi/src/bidi_info.rs
@@ -18,14 +18,14 @@ use std::ops::Range;
 use unic_ucd_bidi::bidi_class::abbr_names::*;
 use unic_ucd_bidi::BidiClass;
 
-use explicit;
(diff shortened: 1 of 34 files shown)


eyeplum:
Hmmm, looks like the `crate::` is supported by stable and 1.26.0.

behnam:
Oh, I thought `master` is broken. Apparently it's not, but the builds here were failing because of the new code was based on the code before we allow warnings again.

So, let's just `crate::` prefixes and all should be fine now. Right?

behnam:
Uh, prev comment was for https://github.com/open-i18n/rust-unic/pull/227 .

Right, we can't actually do this here yet. Let's wait until we bump up min-rustc-version to what's beta right now.

Thanks for working on it, though, and sorry for my misleading comment. :(

PR #236, opened 2018-11-04
ucd/block: Impl PartialCharProperty for Block

behnam:
Also mark `Block` API as incomplete.

---
This change is [Reviewable](https://reviewable.io/reviews/open-i18n/rust-unic/236)


behnam:
bors: r+

PR #250, opened 2019-01-09
Support for stdin #249

### LINKED ISSUE #249: Support for stdin
This should work:

```sh
cat test.txt | unic-inspector
```

LuoZijun:
Resolve #249

### DIFF
--- a/apps/cli/src/bin/unic-echo.rs
+++ b/apps/cli/src/bin/unic-echo.rs
@@ -14,7 +14,11 @@ extern crate clap;
 #[macro_use]
 extern crate unic_cli;
 
-use std::io::{self, Write};
+
+use std::time;
+use std::thread;
+use std::io::{self, Read, Write};
+use std::sync::{Arc, Mutex};
 
 use clap::{Arg, ErrorKind};
 
@@ -111,6 +115,7 @@ fn run() -> Result<()> {
         .arg(
             Arg::with_name("STRINGS")
                 .multiple(true)
+                .required(false)
                 .help("Input strings (expected valid Unicode)"),
         )
         .arg(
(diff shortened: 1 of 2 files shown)


LuoZijun:
@behnam 

If we support accepting parameters from the basic input, 
the main thread will still blocked when this `$ ./unic-echo ` appears. 

Of course, if everyone thinks this is not going to bother the user,
the block style approach will be very straightforward.
and i am happy to change to a synchronous way.

Use case:

*    `./unic-echo "STRING ..."`
*    `cat hello.txt | ./unic-echo` or `./unic-echo < hello.txt`
*    `./unic-echo`

behnam:
Oh, I think I see what you mean.

I think blocking reading from stdin should only happen if we explicitly want to read from `stdin`.

One common condition for reading from stdin is when there are no (non-optional) arguments passed to the CLI. (Like unix's `echo` without any file-names passed to it.)

Another common option is when the argument is one HYPHEN-MINUS (`"-"`). In this case, the position of the HYPHEN-MINUS can tell us if the stdin text should be placed before or after any other regular argument.

If none of them applies (there's at least one argument passed to the CLI and none of the is `"-"`) the stdin is not even (blocking) opened.

So, all of these cases would result in "Hello world" string to be echoed/inspected:

- `unic-echo "Hello world"` and
   `unic-inspect "Hello world"`

- `echo "Hello world" | unic-echo` and
   `echo "Hello world" | unic-inspect`

- `echo "Hello world" | unic-echo -` and
   `echo "Hello world" | unic-inspect -`

- `echo "ignored" | unic-echo "Hello world"` and
   `echo "ignored" | unic-inspect "Hello world"`

- `echo "world" | unic-echo "Hello " -` and
   `echo "world" | unic-inspect "Hello " -`

What do you think?

LuoZijun:
@behnam

I think that's great.

 BTW: Actually is `--`, not `-` :)

LuoZijun:
@behnam ping ?

PR #251, opened 2019-01-09
Add ucd/name_aliases

eyeplum:
Recently I found myself needing to fetch name aliases for a given character and it was not implemented in rust-unic yet. So I implemented it and thought it might be good to merge it back.

The implementation parses http://www.unicode.org/Public/11.0.0/ucd/NameAliases.txt and exposes each property as a static function API in the subcrate `unic_ucd_name_aliases`. 

All of the APIs returns a vector of string references because the comments of `NameAliases.txt` mentions:

> Parsers of this data file should take note that the same code point can (and does) occur more than once.

Didn't find an issue for name aliases but #127 might be related.


unic/ucd/src/lib.rs
```diff
@@ -35,6 +35,7 @@ pub use unic_ucd_category as category;
 pub use unic_ucd_hangul as hangul;
 pub use unic_ucd_ident as ident;
 pub use unic_ucd_name as name;
+pub use unic_ucd_name_aliases as name_aliase;
```
dscorbett: `name_aliase` should be `name_aliases`.

eyeplum: Fixed, thanks!

behnam:
Yeah, something like that! Risks!

Nit: I think the second fn need a different name, though, since Rust doesn't support that kind of overloading (iirc).

PR #268, opened 2019-07-10
impl Hash for CharRange

CAD97:
I don't know why we didn't have this already /shrug


CAD97:
AppVeyor failure is spurious and I forget if I have permissions to do this so let me try

bors: r+

